PR #27, opened 2022-11-10
Renamed the namespaces

enoriega:
Went from `clulab.aske.automates` to `ml4ai.skema.text_reading`


kwalcock:
Thanks for the heads up.  I hope IntelliJ or something helped with the refactoring.  I see some resources in a directory `skema_textreading`.  Should that be `skema.text_reading`?

enoriega:
@kwalcock yes, I used IntelliJ to refactor all the files.

I will fix that directory name

kwalcock:
The conf files are probably a little behind and might not even be used.  I see a TextEngine.basePath being referred to even though it is commented out and has a value with the other format.  That might be for a different PR.  The tests passed for me locally.  We'll probably have the server figured out soon.

kwalcock:
It doesn't matter a lot to me, but the package is `org.ml4ai.skema.text_reading` so I thought the resources would be in `org/ml4ai/skema/text_reading`.  Maybe I need to dust off the monitor, though.

adarshp:
@vincentraymond-ua  - would you be able to look into this issue?

adarshp:
@vincentraymond-ua I've updated the webhook setting, and the POST request seems to now be successfully sent, but for some reason I'm not able to access the interface in the browser. I also tried from the terminal using `curl`:

vincentraymond-ua:
Thats interesting, Curl is working fine for me:
` curl https://ci.kraken.sista.arizona.edu '
'<a href="/welcome">See Other</a>.`

This is probably the DNS issue again. Depending on what type of network you are on, sometimes kraken won't resolve properly. This is what Drew said earlier today:

"Dns record still needs to be copied to martha, until then dns lookups will only work half the time."

enoriega:
@vincentraymond-ua @adarshp I think that the dns entries have propagated already. CI is running again. Thanks!

PR #35, opened 2022-12-02
generate graph representations of MathML

ualiangzhang:


skema/skema-rs/mathml/src/expression.rs
```diff
@@ -1,8 +1,17 @@
...
 };
+use petgraph::dot::{Config, Dot};
+use petgraph::graph::Node;
+use petgraph::visit::NodeRef;
+use petgraph::{graph::NodeIndex, Graph};
+use std::cmp::Reverse;
+use std::collections::VecDeque;
+use std::ptr::null_mut;
+
+pub type MathMLGraph<'a> = Graph<String, String>;
```
adarshp, line 14: I don't think this graph has any trace of MathML left in it after you've done your transformations, right? Perhaps you should name it something different? Like `MathExpressionGraph`?

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +23,1094 @@ enum Atom {
...
     Atom(Atom),
-    Expression { op: Operator, args: Vec<Expr> },
+    Expression {
+        op: Vec<Operator>,
+        args: Vec<Expr>,
+        name: String,
+    },
+}
+
+#[derive(Debug, PartialEq, Clone)]
```
adarshp, line 33: Can you add a short docstring for the `PreExp` struct explaining what it is?

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +23,1094 @@ enum Atom {
...
+            }
+            Mn(x) => {
+                if pre.args.len() >= pre.op.len() {
+                    pre.op.push(Operator::Multiply); // deal with the invisible multiply operator
+                }
+                pre.args.push(Expr::Atom(Atom::Number(x.to_string())));
+            }
+            Mo(x) => {
+                match x {
+                    // Operator::Other(_) => { pre.op = x.clone(); }
```
adarshp, line 57: Can you delete all commented-out code?

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +23,1094 @@ enum Atom {
...
+                Expr::Expression { .. } => {
+                    arg.to_graph(&mut g);
+                }
+            }
+        }
+        return g;
+    }
+}
+
+pub fn get_str_len(str: &str) -> usize {
```
adarshp, line 470: Why not simply use the `len` method (https://doc.rust-lang.org/std/primitive.str.html#method.len) instead of implementing your own function to calculate the length of a string slice?

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +23,1094 @@ enum Atom {
...
+
+pub fn remove_paren(str: &mut String) -> &mut String {
+    while redundant_paren(str) {
+        str.remove(str.len() - 1);
+        str.remove(0);
+    }
+    return str;
+}
+
+// if exists, return the node index; if no, add a new node and return the node index
```
adarshp, line 493: Can you change the double slash `//` to a triple slash `///` - that will make it a doc comment: https://doc.rust-lang.org/beta/rust-by-example/meta/doc.html

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +23,1094 @@ enum Atom {
...
+        op: Vec::<Operator>::new(),
+        args: Vec::<Expr>::new(),
+        name: "".to_string(),
+    };
+    pre_exp.op.push(Operator::Other("root".to_string()));
+    math_expression.to_expr(&mut pre_exp);
+    pre_exp.group_expr();
+    pre_exp.get_names();
+    let g = pre_exp.to_graph();
+    println!("{}", Dot::new(&g));
```
adarshp, line 1098: Can you remove the `println!` calls in the test functions? They are fine for debugging purposes, but once you know the test passes, we don't need them.

adarshp (changes requested):
@ualiangzhang  - I've added a few comments, can you address them? Also, three of the tests are failing (`test_to_expr7`, `test_to_expr8`, `test_to_expr9`). Can you fix them?

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Thanks! I updated the name as you suggested.

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Sure! This is an Intermediate data structure to support the generation of graphs of mathematical expressions.

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: No problem!

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Calling the len() function directly will return the incorrect number of characters in a string. I've been finding and debugging this issue for a long time and had to write my own function.

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Sure!

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Got it! I will remove them.

ualiangzhang:
I noticed that I had changed the way the names were generated, which caused a few tests to fail. But I'll fix them. I also realize that I need to implement something to handle the "equals" operator.

skema/skema-rs/mathml/src/expression.rs
adarshp: @ualiangzhang  Are you sure? Can you provide a minimum working example? 

I just tried changing `get_str_len(name)` to `name.len()` on line 252 and `get_str_len(str)` to `str.len()` on line 449 of `expression.rs`, and the tests are still passing.

Also, could you change the function arguments named `str` to `string` (since `str` is a Rust builtin, it would be good to avoid using it as a variable name - for one, it messes with my syntax highlighting :) )

skema/skema-rs/mathml/src/expression.rs
adarshp: Thanks @ualiangzhang  - I see that you've removed the printing of the DOT output, but there are still some statements like `println!("Success!")` in the tests - could you remove them as well? Running `cargo test` will show us whether tests are successful, so we don't need to additionally print another success indicator.

skema/skema-rs/mathml/src/expression.rs
adarshp: Can you add the explanation as a doc comment above the struct?

adarshp: Sorry, just saw that it is there, nevermind!

skema/skema-rs/mathml/src/expression.rs
```diff
@@ -14,42 +25,1172 @@ enum Atom {
...
+            Mi(x) => {
+                if pre.args.len() >= pre.op.len() {
+                    /// deal with the invisible multiply operator
+                    pre.op.push(Operator::Multiply);
+                }
+                pre.args.push(Expr::Atom(Atom::Identifier(x.to_string())));
+            }
+            Mn(x) => {
+                if pre.args.len() >= pre.op.len() {
+                    /// deal with the invisible multiply operator
```
adarshp, line 55: Doc comments (starting with triple slashes `///`) are not meant to be in the interior of a function - they are the Rust equivalent of Python docstrings, and should only be there at the top of a struct/function/method/module.

For regular comments inside a function, you can use a double slash (`//`)

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: Got it! Will remove them.

skema/skema-rs/mathml/src/expression.rs
ualiangzhang: I see. Thanks for letting me know! I will replace them.

Issue #41, opened 2022-12-07
Issue with artifactory instance


adarshp:
Hi @kwalcock - we are getting some CI build errors that look related to your artifactory instance. Do you know what might be going on? I've attached the error log.

[logs_ml4ai_skema_189_2_2.log](https://github.com/ml4ai/skema/files/10173272/logs_ml4ai_skema_189_2_2.log)

enoriega:
@adarshp there is a scheduled power outage at GS which is affecting all hosted services. Including artifactory.

kwalcock:
@enoriega, I haven't received an all clear yet or any news, but it looks like the machine in question did not go down.  artifactory, aheadaz, elasticsearch, etc. are still running and the machine didn't reboot.  I'm looking at the log next.

kwalcock:
I deleted the file in question on my computer and reran sbt without it complaining.  The checksum matched both before and after.  Do you have a way of manually running sha1sum on the file

`/root/.cache/coursier/v1/http/artifactory.cs.arizona.edu%3A8081/artifactory/sbt-release/org/clulab/model-streamed-trigram-ser_2.12/1.0.0/model-streamed-trigram-ser_2.12-1.0.0.jar`?

The file should have 2,239,559,276 bytes and it is reported to have been downloaded 209 times from [http://artifactory.cs.arizona.edu:8081/artifactory/webapp/#/artifacts/browse/tree/General/sbt-release-local/org/clulab/model-streamed-trigram-ser_2.12/1.0.0/model-streamed-trigram-ser_2.12-1.0.0.jar](artifactory).

I don't know what might have gone wrong except a temporary outage or maybe a limit imposed on file sizes at some point in the process on the CI server.

adarshp:
@kwalcock, @enoriega  - thanks for looking into it! 

@kwalcock  - there is probably a way to manually run sha1sum on that file, but I don't know how to do it off the top of my head. For now I'll leave it be, but if the issue persists, I can take a stab at it.

cl4yton:
@adarshp and @kwalcock : What is the status of this issue? If no longer relevant, close?

PR #47, opened 2022-12-08
Demo updates

adarshp:
This PR
- updates the notebook with more consistent cells using `BASE_URL`, and also updates the docs location from `/api-docs/` to `/docs/` in anticipation of someday deploying this API publicly.
- adds a command-line option to choose the MemgraphDB host (it was previously hardcoded to `localhost`).
- Fixes the `docker-compose.yml` file in the `skema-rs` directory to deal with Docker networking issues.


adarshp:
I'm going to merge this in since the Rust tests are passing. Still not sure what's going on with the checksum mismatch that is causing the TR tests to fail (see #41 ) - but we can look into that later.

PR #55, opened 2023-01-01
Take another pass (or ten) over mention linking

kwalcock:
Here's a draft showing how I understand the mention linking code.  AFAICT it produces the same output file as the previous version does.  Further changes would probably start to deviate from the original, so this is a good time to take stock.  This version is using a debug time stamper and uid stamper so that the output stays consistent.  I recall the hash function for the uid strangely producing different output on the same input.  There are several TODOS:

* escape the names
* add line numbers to comments that don't have them
* adjust dynamically for programming language
* actually deserialize all the json
* update the Scala version :-)
* improve the alignment


enoriega (approved):
All of these changes look great. I agree that it is time of merging before continuing

PR #56, opened 2023-01-04
Show how to use Pdf2Txt

kwalcock:
Output is

```
   rawText: fi gures
cookedText: figures .

   rawText: o ffi cials
cookedText: officials .

   rawText: di ffi cult
cookedText: difficult.

   rawText: sta ffi ng
cookedText: staffing .

   rawText: shif ts
cookedText: shifts .

   rawText: speci fi cally
cookedText: specifically .

   rawText: ca not
cookedText: canot .

   rawText: we ' ve
cookedText: we 've .

   rawText: tra ffi c
cookedText: traffic .

   rawText: fi rst
cookedText: first .

   rawText: Here 's
cookedText: Here 's.

   rawText: Let 's
cookedText: Let 's.

   rawText: hospital s
cookedText: hospitals .
```

The apostrophes are not corrected and they might need a special case.  hospitals and shifts are corrected by the WordBreakBySpacePreprocessor, which is not normally used.  It often overcorrects, as it does here for canot.  I'm not sure that `shif ts` is a normal ligature problem.  Usually the f and t would be together, but it could be added to the list.  It would probably be better to have the mistakes accompanied by their contexts.

### DIFF
--- a/skema/text_reading/text_reading/build.sbt
+++ b/skema/text_reading/text_reading/build.sbt
@@ -13,6 +13,7 @@ libraryDependencies ++= {
   val uJsonVer = "2.0.0"
 
   Seq(
+    "org.clulab"                 %% "pdf2txt"            % "1.1.3",
     "org.clulab"                 %% "processors-main"    % procVer,
     "org.clulab"                 %% "processors-corenlp" % procVer,
     "ai.lum"                     %% "common"             % "0.0.10",
@@ -24,7 +25,7 @@ libraryDependencies ++= {
     "com.lihaoyi"                %% "requests"           % "0.7.1",
     "com.typesafe.play"          %% "play-json"          % "2.9.3",
     "org.scala-lang.modules"     %% "scala-xml"          % "1.0.6", // 2.1.0",
-    "org.clulab"                 %  "glove-840b-300d"    % "0.1.0" % Test,
+    "org.clulab"                  % "glove-840b-300d"    % "0.1.0" % Test,
     "org.scalatest"              %% "scalatest"          % "3.0.9" % Test
   )
 }
--- a/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/apps/Pdf2TxtApp.scala
+++ b/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/apps/Pdf2TxtApp.scala
@@ -1,0 +1,46 @@
+package org.ml4ai.skema.text_reading.apps
+
+import org.clulab.pdf2txt.Pdf2txt
+import org.clulab.pdf2txt.common.pdf.TextConverter
+import org.clulab.pdf2txt.languageModel.GigawordLanguageModel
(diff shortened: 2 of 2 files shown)


adarshp:
@kwalcock  - Unfortunately, I don't have the bandwidth currently to debug the checksum issue. I have a couple of suggestions:

1. You can have me manually re-run the pipeline. Sometimes this works. For example, I just reran the TR CI pipeline and am getting different errors (see attached log file).
2. You can add additional commands lines in the .drone.yml file (https://github.com/ml4ai/skema/blob/27c4003745f48df113461af30136d1c2b8a84da0/.drone.yml#L25-L27) to try and debug the build. Perhaps a minimal script to reproduce the issue using `curl` and `shasum`?
3. If you have the cycles to look into it, I can give you admin privileges on kraken to experiment with the Drone CI server.

[logs_ml4ai_skema_292_2_2.log](https://github.com/ml4ai/skema/files/10354599/logs_ml4ai_skema_292_2_2.log)

enoriega:
@kwalcock looking at the usage example, does it also applies to arbitrary strings? I.e. a paragraph containing one (or more) of noisy token sequences

kwalcock:
@enoriega, yes, it's meant to be used with on "running" text.  Several of the preprocessors submit text to processors in order to get words tokenized, sentences recognized, and case corrected.  For this PR I just didn't have the context available.

enoriega:
Understood. I will figure out where to plug it in the pipeline

kwalcock:
The dependency was read in the second time.  Still, the tests didn't pass, although I thought they did at one time.  It's not working locally, either, so I will check it out.  I don't think it is related to this particular PR.

enoriega:
@kwalcock I did some tracking down and it appears that the problem is adding `org.clulab.pdf2txt` as dependency.
The only potential conflict at glance is that pdf2txt is using processors `8.5.1` and Skema TR  `8.5.2`. 

Maybe bumping up processors to have the same version in pdf2txt and skema could fix the issue. Maybe it is wishful thinking but if it is not complicating other things, it is worth a shot.

kwalcock:
@enoriega, thanks for the tip.  I shouldn't have assumed it was some kind of pre-existing problem.  There seems to be an issue with sbt.  I reported it earlier (https://github.com/sbt/sbt/issues/7090).  For the time being, this work-around is what I could come up with.

enoriega:
Thanks @kwalcock 
@sushmaakoju, I think it is now safe to go and add the change for #65

enoriega:
@kwalcock There seems to be an issue fetching the glove model from maven central. Could it be a result of the recent change in the URLs?

kwalcock:
Both URLs are still available.  This last error is different than those before it.  I'll bump it again.

kwalcock:
The third time was a charm.  I'm not sure what the problem was.  Perhaps there was some network maintenance related to the start of classes.

kwalcock:
I have started to observe network issues related to the download of the large resource myself.  I have asked the department's technical person to look into it.

Issue #54, opened 2022-12-30
It would be useful to have line numbers for docstrings


kwalcock:
Aside from it just not being efficient to throw information away, not having line numbers for docstrings but having them for comments necessitates handling the two differently in code.  It would be great to have access to the docstring line numbers in the files with extracted comments.

adarshp:
@kwalcock - docstrings span multiple lines - would you like just the starting line number, start and end line numbers, or line numbers for each line of the docstring? Even better, if you can provide an example JSON fragment with your preferred format, it will help speed up development.

kwalcock:
Whatever is easiest.  If it's all the same, then something like this below would work fine.  I'm a little concerned that method overloading is not taken into account.  `get_beta` might not be unique.

```json
	"docstrings": {
		"get_beta": {
                    "line_begin": 1,
                    "strings": [
		        "",
		        "    Calculates a rate of exposure given an intrinsic growth rate for COVID-19",
```

adarshp:
@kwalcock Thanks for providing the example! We'll try to match it if possible.

Re: method overloading - good point. I haven't encountered method overloading in Python so far, so it wasn't on my radar. Perhaps we should return a list of 2-tuples of (function name, docstring), instead of a dictionary? Would that work for you?

kwalcock:
Right.  Python doesn't have function/method overloading and it looks like Fortran uses different names for the methods.  Probably we'll never come across it and I'm not sure I should encourage future proofing.  However, Python does have classes and what would happen if two classes had methods of the same name?  That happens often in code, but perhaps not code related to published papers where it might have been simplified and made procedural.  If there needs to be a string for the function, an int for line number, and array of strings for the docstrings, I would favor a dictionary over a tuple.  Does Open API handle tuples at all?  (Apparently it does by using lists of fixed sizes.)  Perhaps the class or module needs to be added in the future to distinguish duplicates so that the tuple becomes unwieldy.  Do the syntax trees not include comments so that comment extraction needs to be done separately like this?

adarshp:
@kwalcock  - @jastier  is working on handling the case where there are two classes with methods of the same name. Basically, we were thinking there would be keys in the dictionary like `"classA.methodA"`, `"classB.methodA"`. 

[...]

PR #58, opened 2023-01-05
Removed Python docstring extraction from comment extractor

### LINKED ISSUE #31: Add support for extracting docstrings for Python classes and modules.
Currently, the comment extraction library extracts docstrings for Python functions, but not for Python classes and modules. We should update it to support this functionality.

adarshp:
It looks like the Python AST contains docstring information, which is output in the GroMEt JSON representation, so we don't need the comment extractor to output the docstrings.

Closes #31 and #54.

### DIFF
--- a/skema/skema-rs/comment_extraction/src/languages/python.rs
+++ b/skema/skema-rs/comment_extraction/src/languages/python.rs
@@ -8,43 +8,19 @@ use nom::{
 };
 use nom_locate::{position, LocatedSpan};
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
 
 type Span<'a> = LocatedSpan<&'a str>;
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Comment {
+    pub line_number: u32,
+    pub contents: String,
+}
+
 #[derive(Debug, strum_macros::Display, Clone, Serialize, Deserialize)]
 enum Statement {
-    Comment {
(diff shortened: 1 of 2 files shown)


adarshp:
@kwalcock  - I've removed the docstring extraction, and the output JSON from the comment extractor now looks like this:

```json
[
  {
    "line_number": 1,
    "contents": "## ============================================================================="
  },
  {
    "line_number": 2,
    "contents": "## NOTATIONAL CONVENTIONS:"
  }
]
```

Let me know if this works for you.

adarshp:
@enoriega  - here's an example from the Google Drive: https://drive.google.com/file/d/1w_v-_LjWmgS2l-K4nt_N5-PQyuKcDSvb/view?usp=share_link

enoriega:
thanks @adarshp . @kwalcock in case you haven't seen this yet, FYI. This may simplify the comment alignment part

kwalcock:
I wouldn't be in a big hurry to remove it, since the information can simply be ignored in the mention linking code.  I haven't checked the file then Enrique referenced, but I do see `"schema_version": "0.1.5",` and probably need to update something before that can be read because last time I failed.  It's strange that later in the file there is a `"schema_version": "0.1.4"`  I'm not sure that Swagger tools are prepared for that.

adarshp:
Thanks @kwalcock ! In that case I'll just add a note about the limitations of the current tool to the README for that module.

PR #59, opened 2023-01-08
Migrating AutoMATES PA functionality, adding Code2FN service

adarshp:
This PR:

- Migrates the code to function network functionality (for Python code at least) from the ml4ai/automates repository to the ml4ai/skema repository.
- Adds the Code2FN web service (thanks to @vincentraymond-ua for implementing this API).
- Adds the `bucky_v2` code in the `data` directory for testing
- Adds a test for the Code2FN pipeline (`skema/program_analysis/tests/test_code2fn.py`). The test is currently failing since the UUIDs for some of the GroMEt components are randomly generated.
- Updates the top-level `README.md` with the following:
  - a more detailed description of the repository's contents and directory structure,
  - instructions on installing the `skema` Python package to the user's virtual environment
  - pointers to instructions for running the Scala and Rust components.


adarshp:
@kwalcock , @enoriega  - I made a minor change to the `setup.py` file and changed some of the import lines in the `gromet_linker` directory to reflect the migration. Let me know if these changes are ok with you.

enoriega:
@adarshp They look fine to me. I don't think they will have consequences to the functionality of our modules

kwalcock:
@adarshp, thanks for the heads up.  It looks like `setup.py` is gone, so I hope that was intended.

adarshp:
> @adarshp, thanks for the heads up. It looks like `setup.py` is gone, so I hope that was intended.

@kwalcock  - yes, it was intentional! I decided to migrate to using a `pyproject.toml` file (see [here](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)). It should still work the same for installing the package  (`pip install -e .`). I figured we don't need a `setup.py` since we are not trying to support legacy builds. We can add it back if the situation changes...

PR #61, opened 2023-01-09
Refactor grounding code

enoriega:
Changed the grounding code in `HomeController` such that it only works on `TexBoundMention`s that are arguments of events and relations and not on all TB mentions. This saves a lot of computational time.


skema/text_reading/text_reading/grounding/src/main/scala/org/ml4ai/grounding/MiraEmbeddingsGrounder.scala
```diff
@@ -52,14 +58,19 @@ class MiraEmbeddingsGrounder(groundingConcepts:Seq[GroundingConcept], embeddings
...
     val topSimilarities = sortedCosineSimilarities.take(k)
     val topConcepts = topKIndices.map(groundingConcepts)
+
     (topConcepts zip topSimilarities) map {
       case (concept, similarity) => GroundingCandidate(concept, similarity)
     }
   }
 
-  def getNormalizedDistance(text:String, name:String): Float = {
+  private val editDistancesCache = mutable.HashMap[(String, String), Float]()
```
kwalcock, line 67: Cool!  Is there any danger of the cache getting too big?  I notice signs of a webapp that might be running a long time.

skema/text_reading/text_reading/grounding/src/main/scala/org/ml4ai/grounding/MiraEmbeddingsGrounder.scala
```diff
@@ -52,14 +58,19 @@ class MiraEmbeddingsGrounder(groundingConcepts:Seq[GroundingConcept], embeddings
...
+  private def getNormalizedDistance(text:String, name:String): Float = {
+
     def getDistance(source: String, target: String) = MED(source, target, allowSubstitute = false, allowTranspose = false, allowCapitalize = false).getDistance
-    return getDistance(text.toLowerCase(), name.toLowerCase())/text.length().max(name.length())
+
+    editDistancesCache.getOrElseUpdate((text, name), getDistance(text.toLowerCase(), name.toLowerCase())/text.length().max(name.length()))
```
kwalcock, line 72: I wonder whether the `(text, name)` should be lowercased in the key, because it is that way in the value and otherwise duplicates might be stored.

kwalcock (approved):
Thanks for the sneak peek.

skema/text_reading/text_reading/grounding/src/main/scala/org/ml4ai/grounding/MiraEmbeddingsGrounder.scala
enoriega: Yes, it is possible, although unlikely at this point because the service is being used ephemerally. 

I want us to use an LRU cache instead of a vanilla map. Probably something like this:
https://stackoverflow.com/questions/14561845/lrucache-in-scala

skema/text_reading/text_reading/grounding/src/main/scala/org/ml4ai/grounding/MiraEmbeddingsGrounder.scala
enoriega: Indeed, also the order shouldn't be relevant for hashing the tuple. Do you know if there is a class that has a position invariant hash?

kwalcock: I think they'd say that you have to store your own type and provide the hash function for it yourself.  Instead of that, you could just sort (lowercased) text and name.  I like the limited-size LRU.

PR #70, opened 2023-01-12
Remove dependency on regextools

kwalcock:
Also migrate to artifactory.clulab.org


enoriega (approved):
Nice! LGTM

kwalcock:
I think so.  The old version was buggy, so at worst there are just new bugs.

PR #74, opened 2023-01-13
Milestone 5 SKEMA Pull Request

titomeister:
Fixed some minor importing issues with the Python to CAST and CAST to GroMEt generation.
Also updated the 'script_functions.py' file to contain the correct arguments.
This is to allow the pipeline to properly generate the Bucky V2 GroMEt like in the automates repository.


vincentraymond-ua (approved):
@titomeister - Looks good! I can also confirm that the Chime Penn Full model is now processed correctly again.

PR #79, opened 2023-01-17
Refactored the COSMOS ingestion pipeline and added an app to annotate…

enoriega:
… COSMOS' json files from the command line

### DIFF
--- a/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/CosmosTextReadingPipeline.scala
+++ b/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/CosmosTextReadingPipeline.scala
@@ -16,7 +16,7 @@ import org.ml4ai.skema.text_reading.serializer.AutomatesJSONSerializer
 import scala.collection.mutable.ArrayBuffer
 
 
-case class CosmosTextReadingPipeline() extends Logging {
+class CosmosTextReadingPipeline extends Logging {
 
   /**
     * Assigns grounding elements to a mention
@@ -26,10 +26,8 @@ case class CosmosTextReadingPipeline() extends Logging {
   def groundMention(m: Mention): Mention = m match {
     case tbm: TextBoundMention => {
       val isGrounded =
-        tbm.attachments.exists {
-          case _: GroundingAttachment => false
-          case _ => true
-        }
+        tbm.attachments.exists(!_.isInstanceOf[GroundingAttachment])
+
       if (!isGrounded) {
         val topGroundingCandidates = grounder.groundingCandidates(tbm.text).filter {
           case GroundingCandidate(_, score) => score >= groundingAssignmentThreshold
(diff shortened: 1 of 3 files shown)


enoriega:
@kwalcock I hope this looks okay!

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/CosmosTextReadingPipeline.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    // TODO: Make this interpretable
+    val textsAndLocations = loader.loadFile(jsonPath)
+    val textsAndFilenames = textsAndLocations.map(_.split("<::>").slice(0, 2).mkString("<::>"))
+    val locations = textsAndLocations.map(_.split("<::>").takeRight(2).mkString("<::>")) //location = pageNum::blockIdx
+
+    logger.info("started extracting")
+    // extract mentions form each text block
+    val mentions = for (tf <- textsAndFilenames) yield {
+      val Array(rawText, filename) = tf.split("<::>")
```
kwalcock, line 116: So these were split, then put back together, and then split again?  I'd avoid that:

```scala
    val lines = loader.loadFile(jsonPath)
    val records: Seq[(String, String, Array[Int], Array[Int])] = lines.map { line =>
      val fields = line.split("<::>")
      val text = fields(0)
      val filename = fields(1)
      val pageNumbers = fields(2).split(",").map(_.toInt) // index may not be correct
      val blockIndexes = fields(3).split(",").map(_.toInt) // ditto

      (text, filename, pageNumbers, blockIndexes)
    }

    val mentions = for ((rawText, filename, _, _) <- records) yield {
```
and later
```scala
      val pageNum = records(id)._3
      val blockIdx = records(id)._4
```
Even so, a `Record` class would be preferable to numbers in the tuple.  Rather than having the `ArrayBuffer` soon thereafter, `flatMap()` could be used.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/CosmosTextReadingPipeline.scala
```diff
@@ -0,0 +1,157 @@
...
+case class CosmosTextReadingPipeline() extends Logging {
+
+  /**
+    * Assigns grounding elements to a mention
+    * @param m mention to ground
+    * @return instance of the mention with grounding attachment if applicable
+    */
+  def groundMention(m: Mention): Mention = m match {
+    case tbm: TextBoundMention => {
+      val isGrounded =
```
kwalcock, line 28: I'm not sure of this logic.  It seems to be the same as
```scala
val isGrounded = tbm.attachments.exists(!_.isInstanceOf[GroundingAttachment])
```
but that seems strange.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/apps/AnnotateCosmosJsonFiles.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.clulab.utils.Logging
+import org.ml4ai.skema.text_reading.CosmosTextReadingPipeline
+
+import java.io.{File, FileOutputStream, PrintWriter}
+
+object AnnotateCosmosJsonFiles extends App with Logging{
+
+    private lazy val textReadingPipeline = new CosmosTextReadingPipeline
+
+    logger.info(s"Starting process with ${args.length} arguments")
```
kwalcock, line 12: If this was a critical application, it would need some exception handling.

kwalcock (approved):
These suggested changes shouldn't change code behavior, except maybe to make it faster.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/CosmosTextReadingPipeline.scala
kwalcock: Maybe even
```scala
      val Array(pageNumbers, blockIndexes) = fields.takeRight(2).map(_.split(",").map(_.toInt))
```

kwalcock: Or
```scala
      val Array(text, filename, rest @ _*) = line.split("<::>")
      val Array(pageNumbers, blockIndexes) = rest.takeRight(2).map(_.split(",").map(_.toInt)).toArray

      (text, filename, pageNumbers, blockIndexes)
```

enoriega:
@kwalcock Thanks for the review. I changed some of the code with your suggestions and will open a new PR with the remaining after the hackaton

PR #84, opened 2023-01-18
Milestone 5 PR #2

titomeister:
Fixed an additional bug that involved function definitions appearing as boxes in the FN. They should now only appear when they're being called.


vincentraymond-ua (approved):
@titomeister - I can confirm this fixes the issue with duplicated FN boxes

PR #102, opened 2023-01-29
enoriega/issue101

enoriega:
Improve the reliability of the TR entry point


kwalcock:
Sometimes the lazy val is so that the expensive part, in this case the `CosmosTextReadingPipeline`, doesn't need to be constructed if someone enters a bad file or directory name.  They won't have to wait so long to find out that they mistyped the argument.  If that is still a concern, one could initialize the variable only after a check to see if `pathsToAnnotate` is non-empty.

kwalcock:
If this has to work through 1000s of files or more, it might good to make sure the files get closed even when there is an exception in `writer.write`.  An inner try and finally could do that.

enoriega:
thanks @kwalcock, I will update these with a separate PR

PR #106, opened 2023-01-30
Assemble for text reading

kwalcock:
This addition *should* make it possible to produce a fat jar file.  We need to fill in the `mainClass` at the very least.  It should be tested, even if unit tests are passing.  Sometimes people have tried to access files that are no longer available outside the development environment and it only shows up now.  Plus I might not have merged some of files incorrectly.  Also, it might be good not to compress the large resource files as they are added to the jar.  It takes a long time and then they need to be repeatedly decompressed each time the program runs, slowing startup time.  I can add that later after everything is known to work.

### DIFF
--- a/skema/text_reading/text_reading/assembly.sbt
+++ b/skema/text_reading/text_reading/assembly.sbt
@@ -1,0 +1,20 @@
+assembly / aggregate := false
+assembly / assemblyMergeStrategy := {
+  case PathList("CHANGES.md")                              => MergeStrategy.discard
+  case PathList("com", "sun", "istack", _*)                => MergeStrategy.first
+  case PathList("com", "sun", "xml", _*)                   => MergeStrategy.first
+  case PathList("javax", "activation", _*)                 => MergeStrategy.first
+  case PathList("javax", "xml", "bind", _*)                => MergeStrategy.first
+  case PathList("logback.xml")                             => MergeStrategy.last
+  case PathList("META-INF", "versions", _*)                => MergeStrategy.first
+  case PathList("module-info.class")                       => MergeStrategy.discard 
+  case PathList("org", "apache", "commons", "logging", _*) => MergeStrategy.first
+  case PathList("org", "bouncycastle", _*)                 => MergeStrategy.first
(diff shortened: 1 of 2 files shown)


enoriega:
@kwalcock can we set the main class whatever the _Play!_ API web entry point?

Basically let `java -jar Skema-TR.jar` be equivalent to `sbt web app/run`

kwalcock:
That `mainClass` setting is for running a Java application and not the webapp.  For the webapp there are several options, but they are probably all more complicated than just a call to java.  I haven't seen a way to run a webapp from a single jar.  There are examples of some options in the [Eidos webapp](https://github.com/clulab/eidos/tree/master/webapp/Docker) subproject.  One strategy is to just use `sbt` to run the webapp, which may not be feasible.  Another option is to use `sbt webapp/stage`, which will deploy all the necessary files to `target/universal/stage`.  There should be a file `bin/webapp` that can be used to start it up.  That does end up calling java, but does so with the support of lots of other files in nearby directories rather than a single jar.  `sbt dist` (described [here](https://www.playframework.com/documentation/2.8.x/Deploying) will take all those staged files and put them into a zip file for deployment elsewhere.  It will need to be unzipped to get the same thing as the staged version.  I hope one of the options is close to what you need.

enoriega:
It sounds too complicated. In that case, we should make the `AnnotateCosmosJson` the main app instead

kwalcock:
AFAIK, the mainClass is just for convenience.  It's possible to run any other static main contained within the jar.  The choice just depends on the expected usage and I wasn't sure how to fill it in.

enoriega:
Right. Those two are the only entry points in use today, so it make sense to have the cosmos one as entry points.

I may write another for free form text soon.

kwalcock:
@enoriega, I merged main into this and then ran both the assembled and unassembled versions on the recent problematic file without any problems.  Assembly seems to work.  I think this can be merged.

enoriega (approved):

enoriega:
Thanks @kwalcock

PR #107, opened 2023-01-30
Fix bug where CWD is lost

mattprintz:
If code that is unable to be parsed by the AST library is submitted, the program can end up in a state such that the current directory does not exist, causing further evocations to fail.

This should ensure that we always return to the original directory regardless of failure.


adarshp (approved):

adarshp:
Thanks @mattprintz !

Issue #115, opened 2023-02-07
`>` Trips Text Reading


enoriega:
The following text trips the TR pipeline:

```
(i.e., R(0) > 0)
```

If you replace the `>` with a `=` it works just fine. It seems to be a problem deep in the odin engine. 

I attach a sample input json file that triggers the exception when running `org.ml4ai.skema.text_reading.apps.AnnotateCosmosJsonFiles`

[documents_5Febsarsdouble--COSMOS-data.json.zip](https://github.com/ml4ai/skema/files/10680070/documents_5Febsarsdouble--COSMOS-data.json.zip)

kwalcock:
Thanks for tracking down the culprit file!

kwalcock:
It seems to be a problem with https://github.com/ml4ai/skema/blob/72aa358ce413fa8d82463e44bea970796766e0d3/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/OdinActions.scala#L282 where a key is not being found.  It looks like a piece of code in which some assumption is not holding.

kwalcock:
...and I'm working on it.

kwalcock:
The code is incorrect.  In the case of a "variable" argument with looksLikeAnIdentifier() being true, the old Mention is replaced by a new one which cannot be in the original synPaths.  It is being patched.

PR #116, opened 2023-02-09
Fix bug that turned out to be in OdinActions

kwalcock:
See #115


enoriega (approved):
LGTM

kwalcock:
It was intended to be complete.

PR #117, opened 2023-02-09
Scenario Context

enoriega:
This is the first implementation of Scenario Context.

The two main abstractions are: `ContextEngine`, which looks around a window of sentences in order to attach location and time scenario context to relation and event mentions; and `MentionOrderer` which is used to abstract the notion of a linear partial ordering of appearance for the mentions, regardless of the source. This way, we can sort the mentions that come from a Cosmos document, which has pages and blocks within pages, and look around efficiently at the time of context resolution.

### DIFF
--- a/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
+++ b/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
@@ -42,25 +42,22 @@ class LocationContextAttachment(locations:Seq[Mention]) extends AutomatesAttachm
     )
   })
 
-  override def toUJson: Value = ujson.Arr(locations map {
-    l => ujson.Obj(
-      "scenarioLocation" -> l.text
-    )
-  })
+  override def toUJson: Value = ujson.Obj(
+    "scenarioLocation" -> (locations map (_.text)).distinct
+  )
 }
 
-class TimeContextAttachment(locations:Seq[Mention]) extends AutomatesAttachment {
-  override def toJson: JsValue = Json.arr(locations map {
+class TimeContextAttachment(times:Seq[Mention]) extends AutomatesAttachment {
+  override def toJson: JsValue = Json.arr(times map {
     l => Json.obj(
       "scenarioTime" -> l.text
     )
   })
 
-  override def toUJson: Value = ujson.Arr(locations map {
-    l => ujson.Obj(
-      "scenarioTime" -> l.text
-    )
-  })
+
+  override def toUJson: Value = ujson.Obj(
+    "scenarioTime" -> (times map (_.text)).distinct
+  )
 }
 
 class MentionLocationAttachment(val filename: String, val  pageNum: Seq[Int], val blockIdx: Seq[Int], attType: String) extends AutomatesAttachment {
--- a/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
+++ b/skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
@@ -1,6 +1,6 @@
 package org.ml4ai.skema.text_reading.attachments
 
-import org.clulab.odin.Attachment
+import org.clulab.odin.{Attachment, Mention}
 import org.ml4ai.grounding.GroundingCandidate
 import org.ml4ai.skema.text_reading.quantities.Interval
 import play.api.libs.json.{JsValue, Json}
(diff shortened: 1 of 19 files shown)


skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/scenario_context/CosmosOrderer.scala
```diff
@@ -0,0 +1,51 @@
...
+          throw new InvalidOperationException("Mention without MentionLocationAttachment")
+
+        val location = getMentionLocation(m)
+        val page = location.pageNum.head
+        val block = location.blockIdx.head
+        val numSentences = m.document.sentences.length
+        OrderingHelper(page, block, numSentences)
+      }
+    ).distinct.sortBy{
+      case OrderingHelper(p,b,s, _) => (p, b, s)
```
kwalcock, line 32: I'm not sure why numSentences should be part of the sort, but then I don't understand the data structure.  If there are two Mentions on the same page and block, then the one whose document has the most sentences should come first?

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/serializer/AutomatesJSONSerializer.scala
```diff
@@ -173,25 +169,22 @@ object AutomatesJSONSerializer {
       case "MentionLocation" => new MentionLocationAttachment(json("filename").str, json("pageNum").arr.map(_.num.toInt), json("blockIdx").arr.map(_.num.toInt), attType)
       case "DiscontinuousCharOffset" => new DiscontinuousCharOffsetAttachment(json("charOffsets").arr.map(v => (v.arr.head.num.toInt, v.arr.last.num.toInt)), attType)
       case "ParamSetAtt" => new ParamSetAttachment(json("attachedTo").str, attType)
-      case "ParamSettingIntervalAtt" => {
+      case "ParamSettingIntervalAtt" =>
         var inclLower: Option[Boolean] = None
         var inclUpper: Option[Boolean] = None
         if (!json("inclusiveLower").isNull) inclLower = Some(json("inclusiveLower").bool)
```
kwalcock, line 175: It looks like this can be `val inclusiveLowerOpt = json("inclusiveLower").boolOpt`.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/utils/MentionUtils.scala
```diff
@@ -44,7 +44,7 @@ object MentionUtils {
 
   def getMentionsWithoutLocations(texts: Seq[String], file: File, reader: OdinEngine): Seq[Mention] = {
     // this is for science parse
-    val mentions = texts.flatMap(t => reader.extractFromText(t, filename = Some(file.getName)))
+    val mentions = texts.flatMap(t => reader.extractFromText(t, filename = Some(file.getName))._2)
```
kwalcock, line 47: I wish that `extractFromText` returned something other than a tuple.  In other projects we've called it an `AnnotatedDocument`.  Then we don't have to do the by the number _._2 thing.

kwalcock (approved):
The code includes lots of hard work.  Thanks!

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
```diff
@@ -35,7 +35,32 @@ class GroundingAttachment(candidates:Seq[GroundingCandidate]) extends AutomatesA
   })
 }
 
-class MentionLocationAttachment(filename: String, pageNum: Seq[Int], blockIdx: Seq[Int], attType: String) extends AutomatesAttachment {
+class LocationContextAttachment(locations:Seq[Mention]) extends AutomatesAttachment {
+  override def toJson: JsValue = Json.arr(locations map {
```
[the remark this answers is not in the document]
enoriega: There aren't any external projects tapping into our code, so we are safe on that front. Maybe we should brainstorm soon about a systematic way: For example, we could use uJson across the board and write a helper function that translates back and forth from PlayJson to uJason an arbitrary Json object which is only plugged into the controller's handlers at request and response time.

About the change, it was necessary to remove redundancy. So it was deliberate :)

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/scenario_context/CosmosOrderer.scala
enoriega: Num sentences is just carried on to subsequent functional calls. On the case of several mentions, its value would be equal so it doesn't matter which comes first and that's why I am calling distinct to keep just one entry per block.

This was me mostly exercising the functional programming muscle instead of doing it imperatively and maybe more legible.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/utils/MentionUtils.scala
enoriega: Good point. I changed the tuple for a case class

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/scenario_context/ContextEngine.scala
```diff
@@ -0,0 +1,57 @@
...
+
+      val contextAttachments =
+        (start to end).flatMap {
+          sentenceIndex =>
+            val contextMentions = mentionsMap.getOrElse(sentenceIndex, Seq.empty)
+            contextMentions
+        }.groupBy(_.label).map {
+          case (label, mentions) if label == "Location" =>
+            new LocationContextAttachment(mentions)
+          case (label, mentions) if label == "Date" => new TimeContextAttachment(mentions)
```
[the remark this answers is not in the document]
enoriega: Fair enough. I replaces map  with collect, which will omit attachments with typos or other issues that aren't accounted through safe typing.

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/attachments/AutomatesAttachment.scala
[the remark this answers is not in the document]
kwalcock: The size of the board in across the board might be an issue.  Several CLU Lab projects. including processors, have used json4s so that it is already a dependency.  In that sense ujson is the odd one out.  For play webapps there may be a way to convince it that some text is already serialized json so that it can be used from json4s or ujson without needing to be converted first to play format.  Processors has a new webapp, so I may check it out.  This ujson part of skema is still quite small, so I'm not very worried about it yet unless there are big plans ahead.  I have a gromet (de)serializer for Scala and used json4s FWIW.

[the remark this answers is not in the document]
enoriega: I have no special attachment to uJson. Json4S should work too. As long as we commit to one, I am happy.

PR #119, opened 2023-02-10
Adding gromet primitive map to Gromet directory

vincentraymond-ua:
Adds the Gromet primitive_map (primitive_map.yaml) to the Gromet directory. It was previously located in the automates repository and never transitioned to Skema.

Also adds primitives for the following types:
- Graph
- Node
- Edge 
- DataFrame


cl4yton (approved):
This looks great! This appears consistent with my interpretation of the gromet types.

PR #131, opened 2023-02-23
Gauravs

gauravsh0812:
Changes I have done:

1) Added the data generation scripts. I have modified them so anyone can work with them by changing the data_generation_config file. 
2) Added the training part and merged it with the inference part to avoid code duplication. One can train the model and then continue using inference scripts without pulling the trained_model (if doing training and inference on the same machine else can pull the trained_model and put it under the "trained_models" folder as described in the README.md file.)


kwalcock:
@gauravsh0812, because there is a pull request on this branch, each push to the server will provoke unit testing and also notifications to others.  If those 160 rounds of unit tests aren't really needed until closer to when the branch is ready to merge, it might be a good idea to cancel the pull request and resubmit it later.  I hope you are able to run the tests locally if need be.

kwalcock:
Thanks!  Pushes without the PR are fine (for me) and maybe even good, because they do help other people see what you're up to.  Have fun programming!

PR #134, opened 2023-03-07
C/C++ comment extraction

jastier:
## What's New

- C, C++ comment parser written in Rust
- Output as a vector of LocatedSpan objects with character offset, line number, and comment text


skema/skema-rs/comment_extraction/src/bin/extract_comments.rs
```diff
@@ -37,15 +39,23 @@ fn main() {
         if let Some(path) = args.output {
             write(path, comments).expect("Unable to write to file!");
         } else {
-            println!("{comments}");
+            println!("{:?}", comments);
```
adarshp, line 42: @jastier  - Can you change this line back to the way it was earlier? We want to print actual JSON, not the debug output.

skema/skema-rs/comment_extraction/src/bin/extract_comments.rs
```diff
@@ -37,15 +39,23 @@ fn main() {
...
         }
     } else if extension == "py" {
         let comments = get_python_comments(input);
         let comments = serde_json::to_string(&comments).unwrap();
         if let Some(path) = args.output {
             write(path, comments).expect("Unable to write to file!");
         } else {
-            println!("{comments}");
+            println!("{:?}", comments);
```
adarshp, line 50: Same here - we want to print actual JSON to be consumed by downstream programs rather than the debug output.

adarshp (changes requested):

adarshp (changes requested):
Also, can you update the C/C++ comment extraction functionality to output JSON (like the Python comment extractor)

skema/skema-rs/comment_extraction/src/bin/extract_comments.rs
jastier: Done.

adarshp (approved):

PR #139, opened 2023-03-22
Comment extraction from directory

adarshp:
This PR updates the `extract_comments` program to be able to recursively extract comments from a directory (the previous capability to handle a single file is preserved). The `output` positional argument has been removed, the program now just prints to standard output (which can be redirected to a file with the `>` operator).


enoriega (approved):
LGTM!

PR #143, opened 2023-03-26
Add endpoint for groundStringsToMira

kwalcock:
Here's a first pass

### DIFF
--- a/skema/text_reading/text_reading/src/main/resources/application.conf
+++ b/skema/text_reading/text_reading/src/main/resources/application.conf
@@ -153,7 +153,8 @@ grounding {
 
 Grounding {
   // General params
-  engine = MiraWebApi // Could be MiraEmbeddings or MiraWebApi
+  // engine = MiraWebApi // Could be MiraEmbeddings or MiraWebApi
+  engine = miraembeddings
   assignmentThreshold = 0.5
   // Params for MiraEmbeddings
   ontologyPath = /mira_dkg_lexical_pretty.json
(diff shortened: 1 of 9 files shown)


kwalcock:
It is getting out of memory errors.  I can add a .jvmopts or .sbtopts for that.  I'm not at all certain you wanted the configuration to be changed to use that engine.  I miss seeing an OpenAPI specificiation for the API or much of any testing.

enoriega:
Can we try doing it like in the linking project? Perhaps it is big enough of a feature to add a new PR for it.

I am not sure how it its set up. With FastAPI (Python), the path `/docs` redirect to that API testing UI and it is added automatically. To call the actual endpoints, `/api` is used. 

Something that follows that convention, albeit loosely, should be okay.

kwalcock:
The linking project at GitHub is https://github.com/clulab/ConceptAlignment FWIW.

enoriega:
@kwalcock I took care of the merge conflict.
@adarshp Looks like there is a change in the PA sub project failing a test that made it into this branch. Should I go ahead and merge this anyway?

kwalcock:
Thanks, @enoriega.  Another way that we sometimes do it is to overwrite in memory some of the values in the config.
```scala
  val groundingConfig = generalConfig.getConfig("Grounding")
  val miraEmbeddingsGrounder = GrounderFactory.getInstance(groundingConfig, chosenEngine = Some("miraembeddings"))
```
becomes
```scala
  val groundingConfig = generalConfig.getConfig("Grounding")
  val groundingConfigWithMira = config.withValue("engine", ConfigValueFactory.fromAnyRef("miraembeddings"))
  val miraEmbeddingsGrounder = GrounderFactory.getInstance(groundingConfigWithMira)
```
It has its own disadvantages, though.

enoriega:
@kwalcock Thanks! I didn't know about this

adarshp:
> @adarshp Looks like there is a change in the PA sub project failing a test that made it into this branch. Should I go ahead and merge this anyway?

I pushed a fix for this just now, and the PA test passes, so sure, go ahead and merge.

PR #146, opened 2023-03-29
xDD Dockerization

enoriega:
Updated the docker file to build an image which is xDD compliant.
This PR include an improvement on `AnnotateCosmosJson` to handler command line parameters


skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/apps/AnnotateCosmosJsonFiles.scala
```diff
@@ -2,53 +2,84 @@ package org.ml4ai.skema.text_reading.apps
 
 import org.clulab.utils.Logging
 import org.ml4ai.skema.text_reading.CosmosTextReadingPipeline
+import scopt.OParser
```
kwalcock, line 5: Cool!

skema/text_reading/text_reading/src/main/scala/org/ml4ai/skema/text_reading/apps/AnnotateCosmosJsonFiles.scala
```diff
@@ -2,53 +2,84 @@ package org.ml4ai.skema.text_reading.apps
...
+        .text("directory to write the output files to"),
+      // more options here...
+      arg[Seq[File]]("<input file 1>...")
+        .unbounded()
+        .action((x, c) => c.copy(inputFiles = x))
+        .text("files to annotate"),
+    )
+  }
+
+  OParser.parse(parser, args, ArgsConfig()) match {
```
kwalcock, line 35: I didn't try it, but probably with
```scala
OParser.parse(parser, args, ArgsConfig()).foreach { config =>
```
you can remove one tab stop throughout.

skema/text_reading/text_reading/Dockerfile
```diff
@@ -38,31 +38,44 @@ RUN sbt publishLocal
...
+# =============================================================================
+
+# =============================================================================
+# SETUP THE SKEMA TR ENVIRONMENT
+# =============================================================================
+RUN mkdir -p /skema/text_reading
+COPY . /skema/text_reading
+WORKDIR /skema/text_reading
+# Compile the fat jar
+RUN sbt assembly
```
kwalcock, line 56: This will work, of course, especially if the machine building the image does not have sbt installed.  Running assembly in the container tends to be very slow, in my experience, and means that sbt and Scala are required in the image as well.

skema/text_reading/text_reading/Dockerfile
```diff
@@ -38,31 +38,44 @@ RUN sbt publishLocal
...
+
+RUN chmod +x annotate_cosmos_json
+
+
 
 # =============================================================================
-# PREPARE FOR USE
+# ENTRY POINT IS THE  WEB API
 # =============================================================================
 CMD ["sbt", "webapp/run"]
```
kwalcock, line 80: This will mean that the webapp is running in development mode, which might not be a good long-term solution.  I do not see, for example, how the $secret from application.conf is being managed.  Startup times for sbt can sometimes be frustrating.

kwalcock (commented):
It's probably most important to get the image to work at all right now.  Later, one could look at

* https://github.com/clulab/eidos/tree/master/webapp#dockerization
* https://github.com/clulab/eidos/tree/master/wmexchanger/Docker
* https://github.com/clulab/ConceptAlignment#preparing-the-docker-image

for examples of how it has been done previously.  (It can be painful!)

skema/text_reading/text_reading/Dockerfile
enoriega: Yes, it is slow, but it is just at build time. We will be posting the image to docker hub. 

The idea is that by having xDD creating the container from our image, they already have the jar ready and can just run the shell script without having to compile or run sbt.

skema/text_reading/text_reading/Dockerfile
enoriega: Indeed. If there is a way to run the web app without relying on `sbt`, we should call this instead.

enoriega:
@kwalcock the image works! I am only waiting on Ian Ross to confirm it is compliant with xDD usage terms and conditions before moving forward.

kwalcock:
Great!  xDD is starting to sound familiar.  I think there was a habitus image for it.  https://github.com/clulab/habitus/blob/main/docker.sbt has traces.  There may have been some concerns about the permissions.

enoriega:
We're still waiting on Ian to tell us about xDD compliance, but the image is ready and the PR contains some useful features.
Any further adjustments will be addressed on a future pull request.

PR #147, opened 2023-03-31
Provide a swagger interface for REST API

kwalcock:
This only does the /groundStringsToMira and then probably even needs better documentation.  The Grammar Visualizer is still available and the URLs of the endpoints have not moved.  Go to http://localhost:9000/api/skema for the swagger interface.

image


enoriega (approved):
Looks great. Thanks!

enoriega:
We will revisit the API soon to add more end points. For now, let's merge. Thanks!

PR #151, opened 2023-04-05
Fix to doc workflow and make use of deps caching 

myedibleenso:
The previous version of the docs publishing workflow was invalid (it had a stray `needs:` entry from a related workflow).

## Summary of changes

- Remove `needs:` entry
- Use haskell/actions/setup@v2
- Add caching to docs workflow for faster builds
   - Before caching 🐢 : [~30 minutes](https://github.com/ml4ai/skema/actions/runs/4623644132)
   - After caching 🦔 : [~3 minutes](https://github.com/ml4ai/skema/actions/runs/4625146649)
- Use `/home/runner/work/skema/skema` as an alternative to `GITHUB_WORKSPACE`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,26 +7,43 @@ on:
     branches: [main]
 
 jobs:
-  haskell:
+  site:
     name: "Build static site"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest]
-    needs: setup
     steps:
     # Checkout code
     - name: Checkout code
       uses: actions/checkout@v2
 
     - name: Set up Haskell
-      # see https://github.com/actions/setup-haskell
-      uses: actions/setup-haskell@v1.1
+      id: setup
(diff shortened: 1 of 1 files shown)


myedibleenso:
Before caching: [~30 minutes](https://github.com/ml4ai/skema/actions/runs/4623644132)
After caching: [~3 minutes](https://github.com/ml4ai/skema/actions/runs/4625146649)

kwalcock:
I can't say that I understand it, so maybe my approval shouldn't enable merging, but I did read it.  I'm a little concerned about the number of languages that are being used: at least Rust, Python, Scala, and Haskell now.

kwalcock (approved):

myedibleenso:
To be clear, @kwalcock , the Haskell stuff was already here and @adarshp is open to replacing this with something else.  In other projects, I've used a Python-based static site generator.  That might make be a suitable substitution here, but it isn't something we need to do immediately.  Once I get to generating API docs for the various languages in the project (Rust, Scala, and Python), I might revisit the static site generator.  

I'm going to look at code coverage and such first, though.

PR #165, opened 2023-04-14
Gromet v0.1.6 PR

titomeister:
This Pull Request includes changes to the CAST, AnnotatedCAST (AnnCAST), and GroMEt structures to  support the newest version 0.1.6 of GroMEt.

Here's a brief list of the changes 

- Changing Loop/Conditional GroMEt structures 
    - Removed explicit call boxes for the different pieces of Loop/Conditional
    - Removed explicit wiring for Loop/Conditional boxes to BFs within them
    - Replaced explicit wiring with having all the Loop/Conditional boxes store 
    - Renamed the different elements of the Loop/Conditionals
- Removal of Imports at the FN array level
    - FN Array level only contains FNs now
    - Imports have been pushed into the call boxes now
    - Imports have different types depending on what kind of import they are.
- Updated variable environment into full symbol table
    - Keeps track of function identifiers, records, and variables
    - variable symbol table functions identically to the previous variable environment, but is now stored within the full symbol table.
- Removal/Rewrite of unnecessary CAST nodes
- Rewritten BinaryOp and UnaryOp to be a single Operator node
- Updating of CAST and AnnCAST passes to support the changing of CAST nodes. 
- Miscellaneous cleanup of code base
- Porting over of cast_to_agraph.py front-end script from automates repository. Used for visualizing CAST.
- Inclusion/Updating of single_file_ingester.py front-end script from Vincent.
- Renaming of CAST2GrFN to CAST2FN to better reflect the contents of the system.


vincentraymond-ua (approved):
@titomeister - The new structure looks good! I am able to go from Python->Gromet for most of the examples I tested and it appears to be accurate and following the new schema.

Here are the two issues I ran into, that still need to be investigated after the merge :
1. AnnCast LiteralValue issue when processing multi-file models
>  File "C:\Users\vraymond\Desktop\Code\skema\skema\program_analysis\CAST2FN\ann_cast\to_gromet_pass.py", line 1064, in visit_assignment
    parent_gromet_fn.pof[pof_length - i].name = elem.val.name
AttributeError: 'AnnCastLiteralValue' object has no attribute 'val'

2. The second line of the following produces an TypeError. It looks like it may be related to the symbol table, but I can't tell for sure.


`
x = 100
`
`
z = x/(10 + 11)
`

> File "C:\Users\vraymond\Desktop\Code\skema\skema\program_analysis\CAST2FN\ann_cast\to_gromet_pass.py", line 149, in comp_name_nodes
    n2_name,n2_id = retrieve_name_id_pair(n2)
TypeError: cannot unpack non-iterable NoneType object

myedibleenso:
v0.1.6 introduces a breaking change in the Bucky model that will be resolved in a subsequent PR.

PR #167, opened 2023-04-14
TS2CAST: Initial framework for Fortran->CAST generation

vincentraymond-ua:
This PR contains the initial framework for generating CAST from Fortran code using Tree-Sitter as a base. The CAST model imports have been update to support the repository structure updates introduced in #165 

 We are able to fully support CAST generation for the space weather model file cpktkm.F:

Currently supported idioms:
- Basic continuation lines
- Integer, Real, Logical, Character variable declaration
- Integer, Real, Logical, Character literal value creation
- Variable assignment
- Single and Multiple dimension array creation, get, set, and slice
- Subroutine and Function definition
- Subroutine and Function calls
- Primitive operators
- Do loop

Additional WIP idioms:
- If/elseif/else conditional


skema/program_analysis/TS2CAST/node_helper.py
```diff
@@ -0,0 +1,102 @@
+from skema.program_analysis.CAST2FN.model.cast import (
```
cl4yton, line 1: Style: Ensure 4 space indentation across code.

skema/program_analysis/TS2CAST/node_helper.py
```diff
@@ -0,0 +1,102 @@
...
+   SourceRef
+)
+class NodeHelper(object):
```
cl4yton, line 4: General question, just to consider: My understanding of the intent here in creating the NodeHelper is to walk the tree-sitter nodes and then add metadata that makes it easy to define a set of predicates (functions that return True/False to indicate whether a node satisfies some properties) as `is_*`, or to retrieve some information. If I'm following, your current approach create is to walk the tree-sitter tree and build up the dict while bookkeeping data that then supports the predicates and info retrieval. I'm wondering whether the NodeHelper bookkeeping needs to be a separate data structure or if the predicates and info could just be implemented to test/retrieve while walking the tree-sitter tree itself. I'm definitely not saying this is wrong, I'm just not sure I understand why there needs to be a separate creation of a persistent NodeHelper bookkeeping store.

skema/program_analysis/TS2CAST/ts2cast.py
```diff
@@ -0,0 +1,739 @@
+import json
```
cl4yton, line 1: Style: Ensure indent is 4 chars across the file. Ensure single blank line between method definitions

cl4yton (approved):
Hi @vincentraymond-ua, This looks great! I have a number of fairly minor style comments:

- Ensure that indentation is 4 characters throughout.
- Ensure 1 space between method definitions (within a class)

Also a general comment: Please do add concise but informative docstrings explaining main classes, methods and functions. You do have good in-line comments throughout ts2cast.py, which is great. But summary docstrings are important for understanding the intended overall logic.

I think I generally follow all of the logic. I do have one question about the intent of the NodeHelper -- I'm wondering if there needs to be separate, persistent bookkeeping, or whether the predicates (is_*) and info gathering methods could instead directly retrieve from the existing tree-sitter node(s) on-demand. But this is not a big issue.

Aside from those minor comments, I think it's ready to merge!

vincentraymond-ua:
Hi @cl4yton - Thank you for your detailed response! I agree with your suggestions, I normally would have fixed the formatting and documentation issues before the pull request, but since it was last minute I didn't quite have time to get to everything before the end of the day. 

As for the NodeHelper class, it was something I created in the very early testing stages. I do agree that this structure can likely be changed, perhaps as a series of helper functions. The current uses for the persistent state are the following:
- Generate the CAST SourceRef and node identifier for each node at the start of execution. 
- Filter out the comments and control characters from a nodes children into separate structures. With these included, it becomes impossible to guarantee that a particular child will correspond to a particular node type.

skema/program_analysis/TS2CAST/node_helper.py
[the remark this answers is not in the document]
vincentraymond-ua: Good to know, I didn't realize this!

skema/program_analysis/TS2CAST/ts2cast.py
```diff
@@ -0,0 +1,812 @@
...
+        self.module.body = []
+        for child in root["children"]:
+            child_cast = self.visit(child)
+            if isinstance(child_cast, list):
+                self.module.body.extend(child_cast)
+            else:
+                self.module.body.append(child_cast)
+
+    def visit(self, node: dict):
+        match node["type"]:
```
myedibleenso, line 83: @vincentraymond-ua , `match` / `case` is a Python 3.10 language feature.  Unfortunately, the project doesn't yet support versions beyond 3.8.  Can you please convert this to `if` / `else` instead?   On a related note, I recommend converting your local develop environment to use Python 3.8.

myedibleenso (changes requested):
@vincentraymond-ua , given that the project is currently stuck on Python 3.8, we'll need to alter `skema/program_analysis/TS2CAST/ts2cast.py` to utilize `if` / `else` in place of `match` / `case`.

myedibleenso:
@vincentraymond-ua , does this include any changes related to ml4ai/skema#85?

myedibleenso (approved):

Issue #174, opened 2023-04-20
Integrating Fortran2GrometFN end-point to skema server


cl4yton:
We now have a programmatic front-end to the Code2FN pipeline for taking Fortran files and processing through to FN (single and multi-file GrometFNModuleCollection):
- Single file: 
- Multi-file: 

Vincent will complete the instructions and once complete will assign to Adarsh to verify it is working with the skema server.

Vincent: 
The interface to the endpoints remains the same. To test, run the skema_server as normal, and send the following as a payload to the fn-given-filepaths endpoint:

`{
  "files": [
    "test.f95"
  ],
  "blobs": [
    "program test\n integer :: x=10\n end program test"
  ],
  "system_name": "string",
  "root_name": "string"
}`

PR #175, opened 2023-04-21
Integrating Fortran->Groment pipeline into skema_server

vincentraymond-ua:
Adds the tree-sitter-fortran dependency required for Fortran->Gromet generation and updates the skema-py Dockerfile. 

Fortran files can now be passed to the existing endpoints to generate Gromet. You are also able to have a mixed system with both Fortran and Python files.

Resolves issue #174


skema/program_analysis/TS2CAST/build_tree-sitter-fortran.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Set working directory to TS2CAST
+wd = os.getcwd()
+os.chdir(root_path)
+
+subprocess.run(["git", "clone", FORTRAN_CLONE_URL])
+
+Language.build_library(
+    # Store the library in the `build` directory
+    "build/my-languages.so",
```
myedibleenso, line 21: how about `os.path.join` or a `pathlib` equivalent here?

skema/program_analysis/TS2CAST/ts2cast.py
```diff
@@ -32,7 +33,8 @@
 class TS2CAST(object):
     def __init__(self, source_file_path: str):
         # Initialize tree-sitter
-        self.tree_sitter_fortran = Language("build/my-languages.so", "fortran")
+        tree_sitter_fortran_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "build/my-languages.so")
```
myedibleenso, line 36: For `build/my-languages.so`, we should create the path programmatically for OS interoperability.

skema/program_analysis/TS2CAST/build_tree-sitter-fortran.py
myedibleenso: Since this subpath is used in two places, can we define and reference a constant for this?

myedibleenso (changes requested):

skema/program_analysis/TS2CAST/ts2cast.py
vincentraymond-ua: Since its wrapped in an os.path.join call, the relative path `build/my-languages.so` should be properly interpreted. However, I do like your idea of defining this somewhere as a constant. Then we could explicitly add some kind of wrapping to it to make sure its OS independent.

vincentraymond-ua: Though maybe not necessarily on second look. I think `os.path.join("build", "my-languages.so")` is what we are looking for here.

adarshp (approved):
LGTM

Issue #177, opened 2023-04-25
Commit SW Fortran source files for testing and demos


myedibleenso:
We should commit supported TIEGCM source code for testing and demos.

- [ ] Add `cpktkm.f` and `cons.f` under `data/TIEGCM`

PR #178, opened 2023-04-25
Adding currently supportd space weather files from TIE_GCM model

vincentraymond-ua:
Adding `cons.F` and `cpktkm.F` Fortran source files from the TIE_GCM space weather model to the data directory.

Note: For some reason, Github's syntax highlighter is having issues with these files, but they are valid and parsable Fortran files.  

Resolves issue #177

### DIFF
--- a/data/TIE_GCM/cons.F
+++ b/data/TIE_GCM/cons.F
@@ -1,0 +1,115 @@
+
+program cons_module
+!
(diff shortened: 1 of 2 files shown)


myedibleenso (approved):
Thanks, @vincentraymond-ua !

PR #180, opened 2023-05-01
Document the API returning mentions and documents

kwalcock:


kwalcock:
I'm going to do it differently so fewer files are marked as changed.

PR #182, opened 2023-05-01
Update text-reading API docs

kwalcock:


kwalcock:
In cases where a description was available from other API documentation, I copied it over.  Where not and it seemed like a description might be useful, I added a todo.  I'm not sure whether it is worth the effort of filling them in.  It may be that someone more familiar with the project can do that much better.

PR #184, opened 2023-05-05
ACSet endpoint and MORAE updates

Free-Quarks:
This updates the ACSet endpoint for the mathml equations to PN to the new model representation json spec. 

This also includes some preliminary work on the MORAE module and the code to model representation pipeline, however they're not yet fully implemented.


adarshp:
Closing this PR for now as @Free-Quarks needs to make some additional updates.

PR #189, opened 2023-05-07
Added a fix to skip empty cosmos files instead of crashing

enoriega:
This is a small fix for an issue reported by the cosmos team when running our image on their side.


enoriega:
Pretty small change, so I went ahead and merged

PR #211, opened 2023-06-01
[CODE2FN] single_file_ingester fixes and snippet_ingester creation 

### LINKED ISSUE #202: [python] Filename truncation in GroMEt generation
Some Python files (TODO: name some specific examples), when sent through the pipeline, have their output GroMEt names truncated by a letter or two. Investigate what's causing this.

vincentraymond-ua:
Fixes filename truncation bug when running the pipeline with single_file_ingester.py. 

Fixes additional bug in single_file_ingester where the `path` argument was previously optional. 

Creates a new ingestion script snippet_ingester.py to ingest code snippets instead of files. Its two arguments are the source string itself and a file extension associated with the snippet language (We need this to pass down the pipeline so we know which language we are working with). 

For example:
```bash
python snippet_ingester.py "x=2" ".py"
```

Resolves #202

### DIFF
--- a/skema/program_analysis/single_file_ingester.py
+++ b/skema/program_analysis/single_file_ingester.py
@@ -1,16 +1,23 @@
 import argparse
 import tempfile
 import os
+from pathlib import Path
 
 from skema.program_analysis.multi_file_ingester import process_file_system
+from skema.gromet.fn import GrometFNModuleCollection
+def process_file(path: str, write_to_file=False) -> GrometFNModuleCollection:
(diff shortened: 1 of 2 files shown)


myedibleenso (approved):
LGTM.  Thanks, @vincentraymond-ua !  Can you please add a section to the wiki describing this runnable?

vincentraymond-ua:
@myedibleenso - Yep! I'm currently working on a page that will describe all three (single, multi, snippet) frontends.

PR #212, opened 2023-06-02
Dockerize webapp in production mode

kwalcock:


skema/text_reading/text_reading/webapp/conf/application.conf
```diff
@@ -15,6 +15,10 @@ play.filters {
   disabled += "play.filters.csrf.CSRFFilter"
    enabled += "play.filters.csp.CSPFilter"
 }
+
+skema.hostname = "skema.clulab.org"
```
enoriega, line 19: Can this be parameterized for running the web app either locally directly with docker or via a docker-compose file?

skema/text_reading/text_reading/webapp/public/schema/api-v1.yaml
```diff
@@ -7,6 +7,8 @@ info:
     name: Clayton T. Morrison
     email: claytonm@arizona.edu
 servers:
+  - url: https://skema.clulab.org
```
enoriega, line 10: Same as above. The use case will involve that our partners will run this through a docker compose file. Hopefully this is easy to support.

skema/text_reading/text_reading/webapp/conf/application.conf
kwalcock: The next line down, `skema.hostname = ${?SKEMA_HOSTNAME}`, allows the value to be taken from an environment variable.  The clulab.org value is in effect a default.  So, I think this one should be OK.

skema/text_reading/text_reading/webapp/public/schema/api-v1.yaml
kwalcock: > Same as above. The use case will involve that our partners will run this through a docker compose file. Hopefully this is easy to support.

This might be harder to change on the fly.  Hopefully there's a way to specify the current host or get the environment variable into the yaml.

skema/text_reading/text_reading/webapp/conf/application.conf
enoriega: Nice!

skema/text_reading/text_reading/webapp/public/schema/api-v1.yaml
enoriega: Okay, I'm sure we will find a way to deal with this

kwalcock:
It seems to be working.  It is deployed on river and running from an image started with docker-compose.  The image is quite large, but it might be put on dockerhub so that others don't need to build it.

enoriega (approved):

Issue #210, opened 2023-06-01
[TR] Update Cosmos JSON endpoint on the WebApp


enoriega:
We have an endpoint that parses cosmos JSON files, where a path to a JSON file is passed as the request's body.

https://github.com/ml4ai/skema/blob/78a1650b03d0f43a27da23327a83561e25913a55/skema/text_reading/text_reading/webapp/app/controllers/HomeController.scala#L287-L298

This end point requires the following enhancements:

- Instead of receiving __a path__ to a JSON file to process, the body should contain __the contents__ of one or more json files. This could be implemented as an array containing the data of each document to annotate as its items
- Update the method name to use camel case instead of snake case.
- Add documentation to the endpoint signature

kwalcock:
@enoriega, where can I find examples of these cosmos json files?

enoriega:
Here are three: [input_examples.zip](https://github.com/ml4ai/skema/files/11643387/input_examples.zip)
Thanks!

enoriega:
Thanks @kwalcock !

PR #214, opened 2023-06-04
Implement cosmosJsonToMentions for webapp

kwalcock:
It seems to work through the browser and there is also a unit test that indicates as much, but I don't know the details of the file format to trust it for sure.  In the OpenAPI file I've left TODOs for all the descriptions and only used placeholders for the input and output formats.  I'd appreciate pointers to any specification available.  These look like very large structures.

This is for #210.


kwalcock:
P.S.  This is a branch off kwalcock/production which should be looked at first in #212.

enoriega (approved):

PR #215, opened 2023-06-05
Taking input and output paths as arguments.

krngrvr09:
Right now the paths are hardcoded. I am making the script more flexible by taking these as arguments because input and output files are copied in a scratch directory in the ht condor cluster.


enoriega (approved):
Looks great!

PR #236, opened 2023-06-14
Updating Gromet FN Python data-model to 0.1.7

### LINKED ISSUE #208: Extend Gromet 0.1.6 schema to support automatic ingestion
Currently, whenever the Gromet schema is updated, the JSON2GroMEt [ingester ](https://github.com/ml4ai/skema/blob/main/skema/program_analysis/JSON2GroMEt/json2gromet.py )needs to be manually updated as well. This causes the ingester to lag slightly behind the current Gromet version. 

I have created a [new version](https://github.com/ml4ai/skema/blob/vraymond/json2gromet_extension/skema/program_analysis/JSON2GroMEt/json2gromet2.py) of JSON2GroMEt that reads the model files directly and does not need to be updated when the schema changes.

However, due to limitations of the current Gromet schema, it is not as efficient, especially with large models. To support efficient ingestion, we would need each data-model class to have a field representing its type name. For example:

`GrometBoxFunction.type_name = "GrometBoxFunction"`

vincentraymond-ua:
Updates the Gromet FN Python data-model classes to version 0.1.7. The only change between 0.1.6 and 0.1.7 is the addition of the `gromet_type` field, which is a string representing the class of an object. 

This change removes any ambiguity when serializing into JSON. For an example see the below code snippet:

```JSON
{
    "schema": "FN",
    "schema_version": "0.1.6",
    "name": "fun1",
    "modules": ["..."],
    "module_index": [
            "fun1"
     ],
    "executables": ["..."],
    "gromet_type": "GrometFNModuleCollection"
}
```

Resolves #208


titomeister (approved):
The changes look good! The classes have all their GroMEt types now.

cl4yton (approved):
Swagger codegen be generating with swag. The generated code looks consistent with the spec!

PR #246, opened 2023-06-19
Model-Extractions linking

enoriega:
Created the utility to link ASKEM text extractions to AMR models.
This utility works for pettiness and regnets. The extractions should come in canonical form, from our data model repository, which means that it works for both SKEMA and MIT text extractions.

In a few days, this will work for gromet function networks as well


myedibleenso:
Hi, @enoriega .  Thanks for the heads up.  I've got a [big PR pending](https://github.com/ml4ai/skema/pull/247) that includes a refactor or `text_reading` to separate the scala and python sources (I noticed that some of your code depends on other pieces of the module).  I wonder if we can wait to discuss merging this until after that is reviewed, since it includes the python pieces of text reading as part of the `skema` module?

- text reading module: https://github.com/ml4ai/skema/blob/3ebf1b9a60b57246762d934bbcc06525edd65e0d/pyproject.toml#L71-L72

- text reading dependencies (may need to have your pinned versions assuming these are compatible with `img2mml`): https://github.com/ml4ai/skema/blob/3ebf1b9a60b57246762d934bbcc06525edd65e0d/pyproject.toml#L53

^ those are included based on imports in your code.  If any of those can be removed, though, let's prune things.

- text reading script (maybe not necessary?): https://github.com/ml4ai/skema/blob/3ebf1b9a60b57246762d934bbcc06525edd65e0d/pyproject.toml#L85

enoriega:
Yes. We can hold this PR off until #247 is ready.

myedibleenso:
@enoriega , please reopen when ready and conflicts with `main` have been resolved.

PR #248, opened 2023-06-20
Model extractions linking into mkdocs-migration

enoriega:


myedibleenso:
@enoriega , this doesn't seem to remove any code (it only introduces additions), but you'd mentioned pieces would be deprecated.  Would you please update this PR to remove anything from https://github.com/ml4ai/skema/tree/ghp/mkdocs-migration/skema/text_reading that is no longer needed?

PR #255, opened 2023-06-22
ISA: Optimize the seed selection methods in SGM

ualiangzhang:


adarshp (approved):

adarshp (changes requested):
@ualiangzhang  - can you add the Levenshtein package as a dependency in pyproject.toml?

PR #261, opened 2023-06-22
Fixes #260

### LINKED ISSUE #260: [TR] Paralelize TR webservice processing
`cosmosJsonToMentions` works over multiple input files sequentially. Parallelize this function to handle multiple files using multithreading

enoriega:


myedibleenso (approved):
LGTM, @enoriega !  I was going to say that in the future you may want to add some limits to your thread pool size, but I think this can be handled another way when launching your docker container.

enoriega:
Thanks. I'm curious about how to control it through docker

myedibleenso:
>Thanks. I'm curious about how to control it through docker

Untested, but I *think* specifying the number of CPUs via `--cpus=` will set an upper limit on your default thread pool.

PR #267, opened 2023-06-26
[code2fn] - Adding zip->Gromet endpoint

### LINKED ISSUE #231: [code2fn] - Finish .zip -> GrometFNModuleCollection endpoint
We have the base for an endpoint that accepts a source code module as a zip file. We need to update it to support Fortran ingestion and then include it with the rest of the endpoints.

vincentraymond-ua:
- Adds new endpoint fn-given-filepaths-zip that processes a code system stored in a zip archive and returns Gromet JSON. The following is an example call to the endpoint:

```python
file = {
    "zip_file":open("code_system.zip", "rb")
}
response = requests.post(URL, files=file)
```

- Updates fn-given-filepaths endpoint to make `system_name` and `root_name` fields optional.

- Rewrites parts of fn-given-filepaths to use pathlib instead of os.path.

Resolves #231


myedibleenso (changes requested):
Great, @vincentraymond-ua !  Following what we did for https://github.com/ml4ai/skema/blob/414dcceaad59af056b312056b21d6d392254d213/skema/img2mml/eq2mml.py#L75-L83, can you please add the Python code snippet in your PR to the docstring for `root(` ?  Maybe you can prefix your snippet with a demo of how to create a zip archive from a code directory as well.

skema/skema_py/server.py
```diff
@@ -39,31 +79,67 @@ def ping():
...
-        # Create system_filepaths.txt file
-        system_filepaths = os.path.join(tmp, "system_filepaths.txt")
-        with open(system_filepaths, "w") as f:
-            f.writelines(file + "\n" for file in system.files)
+    ### Python example
+    ```
+    import requests
 
-        ## Run pipeline
-        gromet_collection = process_file_system(
```
myedibleenso, line 58: @vincentraymond-ua , can you please include this import as well?

```python
from skema.program_analysis.multi_file_ingester import process_file_system
```

myedibleenso (approved):
Thanks, @vincentraymond-ua !

PR #271, opened 2023-06-26
Added endpoint for extracting comments from zipfile

adarshp:
This PR adds a REST API endpoint to the skema-rs service for extracting comments from a zipfile.

Example curl invocation:

```
curl -X POST localhost:8080/get-comments-from-zipfile --data-binary "@data.zip"
```

### DIFF
--- a/skema/skema-rs/Cargo.lock
+++ b/skema/skema-rs/Cargo.lock
@@ -49,13 +49,13 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand",
+ "rand 0.8.5",
  "sha1",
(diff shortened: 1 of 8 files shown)


myedibleenso (approved):
LGTM.  Thanks, @adarshp , for adding that endpoint and modifying the previous one to support multiple languages.

adarshp:
@myedibleenso  - clarification - the `extract-comments` endpoint still assumes Python. The `extract-comments-from-zipfile` endpoint can deal with a zipfile containing a mix of python, fortran, C/C++ files.

If you need the single-file endpoint to support multiple languages for the demo, let me know.

PR #279, opened 2023-06-27
Update Dockerfile.skema-py

enoriega:
Added a `pip install` dependency required to support running the TR notebook

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ authors = [
 description = "Scientific Knowledge Extraction and Model Analysis"
 version = "0.1.0"
 dependencies=[
-    "gensim",
+    #"gensim",
     "dill",
     "networkx",
     "PyYAML",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ isa = [
 img2mml = ["fastapi", "uvicorn", "torch", "torchvision", "python-multipart"]
 
 # dependencies for text reading utilities.
-tr = ["transformers", "gensim", "SPARQLWrapper", "tqdm", "pandas"]
+tr = ["transformers", "SPARQLWrapper", "tqdm", "pandas", "askem_extractions[all]@git+https://github.com/ml4ai/ASKEM-TA1-DataModel"]
 
 # project documentation generation
 doc = ["mkdocs==1.2.3", "pdoc3==0.10.0", "mkdocs-git-snippet==0.1.1", "mkdocs-git-revision-date-localized-plugin==0.11.1", "mkdocs-git-authors-plugin==0.6.3", "mkdocs-rtd-dropdown==1.0.2", "jinja2<3.1.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,6 @@ authors = [
 description = "Scientific Knowledge Extraction and Model Analysis"
 version = "0.1.0"
 dependencies=[
-    #"gensim",
     "dill",
     "networkx",
     "PyYAML",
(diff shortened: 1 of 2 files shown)


myedibleenso:
@enoriega , this can't be included for the time being (next couple of months), as your library depends on Python >=3.9.  `skema` is currently stuck on v3.8 until we move to using `tree-sitter` to process Python source code for program analysis.

Once that change has been made, it would be preferable to add this dependency under [the `tr` optional dependencies in the `pyproject.toml`](https://github.com/ml4ai/skema/blob/9b7c26c165c149254901cf1feba089b54a0b458f/pyproject.toml#L57) (ex something like `"askem_extractions[all]@git+https://github.com/ml4ai/ASKEM-TA1-DataModel"`)

myedibleenso:
For the time being, I suggest using a separate example subdirectory with its own docker-compose.yml file specific to text reading in the DockerVM repo.  Will that work?

enoriega:
Yeah. I think this works for tomorrow. The code should run with python 3.8, albeit a few minor type hint changes. I will make sure it works and follow your advice on the dependency install

myedibleenso:
[Here is a recap of the discussion from 4/6/2023 on Slack about this](https://mlbayes.slack.com/archives/CF1S834KH/p1680836252866099?thread_ts=1680820645.268039&cid=CF1S834KH):

>GHP: I noticed in the pyproject.toml that the python library does not currently work with Python versions > 3.8 (presumably due to something about the program_analysis package).  Does this relate to a particular dependency?  I don’t see any specific deps (https://github.com/ml4ai/skema/blob/84f804005f3227f4826c67cbdbe6933517273dfe/pyproject.toml#L12-L16) that would make me think so.  I ask partly because I was able to install everything using 3.9 (after changing the toml) and run the tests (two failing).

>CM: There is a current dependency of the PA pipeline on Python 3.8 b/c it depends on the python native ast. The ast, unfortunately, does not guarantee backward compatibility, so changes in 3.9 break things. This is something that in the future I hope we can remove, perhaps by migrating the python front-end to use tree-sitter’s python parser. But right now there are higher priorities.

enoriega:
Right. `gensim` was the culprit that pinned me to 3.9 at least for TR. Now that I am reworking the gromet aligner, we are replacing gensim with huggingface and I think it should work. So maybe soon I can make all the TR python code compatible with 3.8.

myedibleenso (changes requested):
- remove TR util library's dependency on Python v3.9
- add dependency under `tr` optionals in `pyproject.toml` (instead of listing in Dockerfile)

pyproject.toml
```diff
@@ -11,7 +11,7 @@ authors = [
 description = "Scientific Knowledge Extraction and Model Analysis"
 version = "0.1.0"
 dependencies=[
-    "gensim",
+    #"gensim",
```
myedibleenso, line 14: Let's just remove this.

pyproject.toml
```diff
@@ -54,7 +54,7 @@ isa = [
 img2mml = ["fastapi", "uvicorn", "torch", "torchvision", "python-multipart"]
 
 # dependencies for text reading utilities.
-tr = ["transformers", "gensim", "SPARQLWrapper", "tqdm", "pandas"]
+tr = ["transformers", "SPARQLWrapper", "tqdm", "pandas", "askem_extractions[all]@git+https://github.com/ml4ai/ASKEM-TA1-DataModel"]
```
myedibleenso, line 57: Awesome!  Have you confirmed (ex. from the REPL) this pulls in your dep?

myedibleenso (changes requested):

pyproject.toml
enoriega: Yes, it works. It builds and I am able to import the code from within a python session inside the container

myedibleenso (approved):

PR #314, opened 2023-07-07
Unary Operator Wire Fix PR

### LINKED ISSUE #200: [python] Incorrect Wiring in CHIME Penn model
Justin pointed out some wires in the CHIME_penn_full model have -1s for their "src" and "tgt" fields when they should have the correct index. This may be related to the guards placed by the Bucky model.

### LINKED ISSUE #268: Wiring bug for gromet FN
There seems to be a small wiring bug for some gromets in version 0.1.7. Some variables were not wired into their expressions correctly. (perhaps related to USub primitives, but not sure)

Worry about after openhouse though

titomeister:
This PR addresses an issue with wiring when Unary operations (i.e. UnarySub/Minus and the rarer UnaryAdd). The wires were not being generated correctly and were returning negative indexing. This has now been fixed with a rewrite of handling the Unary Operations.

This PR also introduces some initial tests for the expression files that check properties and correctness of the Gromets.

Resolves #200 
Resolves #268


skema/program_analysis/CAST2FN/ann_cast/to_gromet_pass.py
```diff
@@ -930,7 +924,6 @@ def determine_func_type(self, node):
                     # TODO: import_version/import_source
                     return (FunctionType.IMPORTED, ImportType.OTHER, None, None, "Python", "3.10")
```
myedibleenso, line 925: I see 3.10 here.  Can you explain how this relates (or doesn't relate) to skema being stuck on 3.8?

myedibleenso (approved):
Thanks, @titomeister .  🦄  for the tests.

I'd like @vincentraymond-ua to also review this, but LGTM. 

One question for you.  I see a few lines like the following:

```python
return (FunctionType.IMPORTED, ImportType.OTHER, None, None, "Python", "3.10")
```

These lines appear to reference Python 3.10.  Can you explain how this relates (or doesn't relate) to skema being stuck on Python 3.8?

skema/program_analysis/CAST2FN/ann_cast/to_gromet_pass.py
vincentraymond-ua: I'll let @titomeister answer for confirmation, but I think in this case the 3.10 is referring to the target source language version, not the source language of skema. Though I'm curious how that function would behave if passed Fortran code instead of Python.

vincentraymond-ua (commented):
@titomeister - Overall LGTM! My only recommendation would be to run `black` on both Python files for formatting.

skema/program_analysis/CAST2FN/ann_cast/to_gromet_pass.py
titomeister: I think I just assumed that any code written nowadays would be Python 3.10 (or whatever the latest version is now as that was a little while ago). It doesn't have any real relation to us being stuck on 3.8.
The reason we're stuck on 3.8 has to do when we ingest a file and generate its Python AST, the Python AST that gets generated has changed from version 3.8 and the handlers aren't yet setup correctly to handle the AST from later versions of Python. If I recall correctly it's only the handler that has to do with indexing that needs some changes.

I'm not sure how the function would behave with Fortran code, as I think I wrote this before we started supporting modules in Fortran. It probably needs some changes to appropriately generate the correct Gromet for Fortran.

myedibleenso:
Thanks, @titomeister !

PR #315, opened 2023-07-07
[integration] endpoint documentation and MML sterilization for use with MORAE

myedibleenso:
## Summary of changes
- **[bugfix]** switch from `opencv-python` to `opencv-python-headless`
   - opencv has many requirements that were unaccounted for in docker image we were building.  We were unknowingly publishing broken builds.  Adding all of those deps would bloat our image size without any real benefit.  With the switch, we pull in only the minimal set of features needed for img2mml.
- **[bugfix]** OpenAPI documentation fixed ("decription" -> "description", etc.)
- **[enhancement]** prune HTML attributes in MML tags before sending to MORAE
- **[bugfix]** healthcheck response_type: Text -> int (status code)
- **[documentation]** LaTeX equation examples using Lotka-Volterra equations
- **[QoL]** default value for mathjax endpoint uses URL for deployment
- **[refactor]** move data models used by `img2mml` endpoints to `img2mml.schema`


myedibleenso:
@ualiangzhang , we should reconsider whether opencv is a necessary dependency for this project as its only use currently is this line in `translate.py`:

```python
inverted = cv2.bitwise_not(image_arr)
```

We should be able to accomplish that in numpy or torch/torchvision.

adarshp:
Just noting that you shouldn't need to prune attributes in the MML tags, they should be handled by the parser.

Issue #311, opened 2023-07-07
[skemy-py] Update fn-given-filepaths-zip to support comment unification


vincentraymond-ua:
Currently only fn-given-filepaths supports comment-unification. Once #310 is complete, we can use the System enrichment function to handle this functionality in fn-given-filepaths.

vincentraymond-ua:
## Update 8/22/2023

Beginning work on switching to tree-sitter comment extractor for comment unification. Should be a relatively simple process.

vincentraymond-ua:
## Update 8/25/2023

Resolved by #266

PR #320, opened 2023-07-10
[skema-py] - Automatically enrich System with comments

### LINKED ISSUE #310: [skema-py] Enrich System with comments
Tasks:
- [ ] Create a utility function that takes a System as input, runs the comment-extractor, and returns the system with comments added.
- [ ] Update the workflows that accept code as input to enrich the System first before extracting Gromet.

vincentraymond-ua:
Adds `system_to_enriched_system` function which takes a System as input and attempts to return an enriched system with comments added. This is done by running the calling the Rust comment extractor proxy endpoint over the files in System.blobs.
**Notes:**
-  Due to a quirk of the fn-unifier, System.root_name must be set for automatic comment enrichment to work. Gromet will still be returned from the endpoint, but the comments will not be included. This only affects the fn-given-filepaths endpoint. fn-given-filepaths-zip will automatically determine the root_name.

Adds a call to `system_to_enriched_system` in `system_to_gromet`. Now any endpoint using this function (fn-given-filepaths, fn-given-filepaths-zip) will automatically attempt to extract and unify comments. 

Removes comment enrichment stubs from skema/rest/workflows.py. This logic will be handled directly from the endpoints.

Resolves #310 #311

### DIFF
--- a/skema/skema_py/server.py
+++ b/skema/skema_py/server.py
@@ -269,8 +269,8 @@ async def fn_given_filepaths_zip(zip_file: UploadFile = File()):
                 blobs.append(zip.open(file).read())
 
     zip_obj = Path(zip_file.filename)
-    system_name = zip_obj.name
-    root_name = zip_obj.name
+    system_name = zip_obj.stem
+    root_name = zip_obj.stem
 
     system = System(
         files=files, blobs=blobs, system_name=system_name, root_name=root_name
--- a/skema/skema_py/server.py
+++ b/skema/skema_py/server.py
@@ -80,7 +80,6 @@ def system_to_enriched_system(system: System) -> System:
 
     # Check if Rust is installed on the system. If not, then return the system as is.
     if not which("cargo"):
-        print("HERE")
         return system
 
     # Create a temporary directory to store the input source files.
--- a/skema/skema_py/server.py
+++ b/skema/skema_py/server.py
@@ -79,7 +79,8 @@ def system_to_enriched_system(system: System) -> System:
     skema_rs_path = Path(__file__).parent.parent / "skema-rs" / "comment_extraction"
 
     # Check if Rust is installed on the system. If not, then return the system as is.
-    if which("cargo") is None:
+    if not which("cargo"):
+        print("HERE")
         return system
 
     # Create a temporary directory to store the input source files.
--- a/skema/skema_py/server.py
+++ b/skema/skema_py/server.py
@@ -73,9 +73,10 @@ class System(BaseModel):
         },
     )
 
+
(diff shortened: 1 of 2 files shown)


skema/skema_py/server.py
```diff
@@ -65,6 +71,46 @@ class System(BaseModel):
...
+    tmp_path = Path(tmp.name)
+    
+    comments = {"files": {}}
+    for file, blob in zip(system.files, system.blobs):
+        file_path = Path(tmp_path, system.root_name or "", file)
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+        file_path.write_text(blob)
+
+        result = subprocess.run(
+            ["cargo", "run", "--", str(file_path)],
```
myedibleenso, line 97: 1. Does this include an invocation of the Rust runnable?

myedibleenso (changes requested):
If we want to go this route (as opposed to proxying REST API requests to the skema-rs piece), we probably want a unified docker image (which means we may want to rename the image, etc).  

@adarshp , what do you think?

skema/skema_py/server.py
vincentraymond-ua: Yes, the full invocation would look like
`cargo run -- /tmp/root_name/example.py`.
**cargo run** is the Rust runnable, **--** specifies that the code to run is in the cwd, and  **/tmp/root_name/example.py** is the path to the source file.

vincentraymond-ua:
7/10/23 4:00 EST - Met with @myedibleenso, decided to convert subprocess call to proxy call to /extract-comments and /extract-comments-from-zipfile endpionts.

adarshp:
@myedibleenso  - I think a unified Docker image is not necessary, I think converting the subprocess call to a proxy call is the right call

skema/skema_py/server.py
```diff
@@ -65,7 +71,38 @@ class System(BaseModel):
...
+        return extension_map.get(extension, "")
+
+    comments = {"files": {}}
+    for file, blob in zip(system.files, system.blobs):
+        file_path = Path(system.root_name or "") / file
+
+        snippet = CodeSnippet(
+            code=blob, language=get_language_from_extension(file_path.suffix)
+        )
+        result = await comments_proxy.proxy_extract_comments(snippet)
```
myedibleenso, line 97: to avoid blocking, we might want to try using `await asyncio.gather()`.

vincentraymond-ua: I've now updated this part of the code to use asyncio.gather() instead.

myedibleenso (approved):
This looks great.  Thank you, @vincentraymond-ua !

PR #328, opened 2023-07-13
TR REST Proxy Calls

enoriega:
Added a few proxy endpoints to the TR router to expose some UWisc, SKEMA and MIT api calls that are satellite to the integrated TR pipeline


cl4yton (approved):
This looks good!

PR #329, opened 2023-07-13
[code2fn] - Support for Fortran Record Idiom and TS2CAST enchancements

vincentraymond-ua:
### **Refactoring pipeline to work directly on tree-sitter nodes:**
Previously,  we were generating our own Dict tree based off of the tree-sitter parse tree. This allowed us to make a few quality of life optimizations such as removing nodes representing control characters like '+' and '=='. 
  
The code has now been refactored to remove this step. This change decreases processing time and will make it easier to support other tree-sitter languages in the future. 

In its place, multiple helper functions have been defined in node_helper.py to better support working with tree-sitter Node objects.

### **Moving tree-sitter parsing out of preprocessor:** 
The tree-sitter parsing of the Fortran source code has been moved out of the preprocessor and into ts2cast.py. Overall, it makes more sense for the preprocessor to return source code rather than a parse tree.

### **Adds supports for Record Idiom:**
We are now support generating CAST for the Record idiom in Fortran. Specifically, the record idiom shows up as a Derived Type.
``` Fortran
type, public :: Circle
   real :: radius
contains
   procedure :: area => circle_area
   procedure :: print => circle_print
end type Circle
```

### **Adds support for multi-module CAST output:**
Previously, there was a 1->1 relationship between source file and CAST module. However, with the addition of the Record idioms, the following are now being processed as their own CAST module:

- Module body
- Program body
- Functions defined outside of module/program

The source->Gromet frontends have also been updated to support this change.

### DIFF
--- a/skema/program_analysis/TS2CAST/node_helper.py
+++ b/skema/program_analysis/TS2CAST/node_helper.py
@@ -4,32 +4,33 @@ from skema.program_analysis.CAST2FN.model.cast import SourceRef
 from tree_sitter import Node
 
 CONTROL_CHARACTERS = [
-            ",",
-            "=",
-            "(",
-            ")",
-            "(/",
-            "/)",
-            ":",
-            "::",
-            "+",
-            "-",
-            "*",
-            "**",
-            "/",
-            ">",
-            "<",
-            "<=",
-            ">=",
-            "only"
+    ",",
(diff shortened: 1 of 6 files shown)


cl4yton (approved):
This looks great!

PR #331, opened 2023-07-14
Fix a bug in retrieve model

ualiangzhang:
If the model path is not existed when calling the retrieve_model function, download the default model from the server.


ualiangzhang:
@myedibleenso Could you please take a quick look? I just added one more condition when retrieving the model, as suggested by Enrique. Thanks!

myedibleenso (approved):
LGTM.

PR #332, opened 2023-07-14
[TR][METAL] Added a router for METAL (which is link_amr for now)

enoriega:
Here will go endpoints to link AMR, Gromets and the future METAL features


enoriega:
Will try to sort out the unit test issues here

cl4yton (approved):
This looks good to me!

PR #343, opened 2023-07-17
Changed the default address for the inner MIT service

enoriega:


enoriega:
@myedibleenso just wanted to verify that the endpoint is using `http` not `https`

myedibleenso:
Maybe you want to deploy that other TR service on UA infra as well?

myedibleenso (changes requested):
As I understand it, I believe the plan is now to update the ENV defaults to use addresses for the TR services deployed on UA infrastructure.

enoriega:
@myedibleenso Yes. We actually end up using MIT's deployment as default value. Thanks!

PR #347, opened 2023-07-17
Hackathon fn2amr

Free-Quarks:
This PR: 
-adds basic (only multiplication) support for the expression string fields in AMR extraction
-updated the fn2amr endpoint to use mathexpressiontree's (only PertriNet)
-improves partial extraction support for all AMR's , (still gives back states and parameters if transitions fail)


cl4yton (approved):
looks good!

PR #349, opened 2023-07-17
Fix OpenAPI docs to include request body examples

enoriega:
Fixed the issue with the documentation. 
For future reference:
The problem was caused by adding a hash function to the`TextReadingDocumentResults` model and calling the hash magic function on an unhashable type (`AttributeCollection`).

I implemented hashing on `AttributeCollection` on the `ASKEM-TA1-DataModel` repo.

The `Skema` package will need to be reinstalled to fetch the updated dependency correctly.

### DIFF
--- a/skema/rest/schema.py
+++ b/skema/rest/schema.py
@@ -122,17 +122,17 @@ class MiraGroundingOutputItem(BaseModel):
     )
     groundingConcept: MiraDKGConcept = Field(
         description="DKG concept associated to the query",
-        # example=MiraDKGConcept(
-        #     id="apollosv:00000233",
-        #     name="infected population",
-        #     description="A population of only infected members of one species.",
-        #     synonyms=[],
-        #     embedding=[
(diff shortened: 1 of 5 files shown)


skema/rest/schema.py
```diff
@@ -81,43 +97,42 @@ class MiraGroundingInputs(BaseModel):
...
-            name="infected population",
-            description="A population of only infected members of one species.",
-            synonyms=[],
-            embedding=[
-                0.01590670458972454,
-                0.03795482963323593,
-                -0.08787763118743896,
-            ]
-        )
+        # example=MiraDKGConcept(
```
myedibleenso, line 125: @enoriega , if we uncomment this, are the docs negatively affected?

enoriega: It shouldn't. If I comment it I can see it in swagger and it doesn't break the docs

image

enoriega:
Merging the change

PR #350, opened 2023-07-17
[code2fn] - Pydantic model_validate fix 

vincentraymond-ua:
Switches from using System.model_validate to System.parse_obj. The former is not compatible with Python version 3.8


myedibleenso (approved):
Hopefully we can leave behind our v3.8 soon...

PR #352, opened 2023-07-17
Lieffers eq2amr bugs

Free-Quarks:
This is a small bug fix for the eq2amr pipeline. It was messing up Unary subtractions when inside a multiplication during the flatten_mults() function call. (It was a small scoping mistake)


cl4yton (approved):
Good!

PR #354, opened 2023-07-18
[rest] - Fixing snippets-to-pn-amr and repo-to-pn-amr

vincentraymond-ua:
There was a bug in the amr workflows caused by using the wrong http request method when communicating with the skema-rs services. This has been fixed by switching these requests from POST to PUT.

Additionally, the regnet endpoints `code_snippets_to_rn_amr` and `repo_to_rn_amr` have been temporarily disabled because they are outdated.


Free-Quarks (approved):
Looks good to me! Thanks Vincent!

PR #372, opened 2023-07-27
Memory map local word embeddings

kwalcock:


kwalcock:
This is configured to work for the MiraEmbeddingsGrounder.  The goal is to use it in other places as well, but this is a good point to check in.

kwalcock:
I'm still going to add a finalizer to this.

kwalcock:
@myedibleenso, you may be interested.

kwalcock:
After reading https://mapdb.org/blog/mmap_files_alloc_and_jvm_crash/, I decided that the finalizer should be skipped.  It says

> There is no official way to close memory mapped buffer. JVM releases file handles only after ByteBuffer gets garbage collected. If GC is not invoked for some time (large free heap), you will run out of file handles.

If my finalizer is being called, then most likely the buffers are about to be closed anyway and the connections are soon to follow.  My mucking around in it will only complicated the situation.  We expect only a handful of memory mapped files anyway.

enoriega (approved):
Really nice job! @kwalcock

PR #374, opened 2023-07-27
Jastier/refactor_CAST2FN

jastier:
Refactored CAST2FN into CAST/Fortran

Please test this code to be sure no linkages have broken.


myedibleenso:
Thanks, @jastier !  Curious what others think: my preference would be to lower case the directories corresponding to frontends (`Fortran` -> `fortran`); maybe even CAST (`CAST` -> `cast`).  @vincentraymond-ua @titomeister @cl4yton ?

jastier:
Sure @myedibleenso , these would not be hard changes to make.   We would end up with something like this:

```
CAST
├── fortran
├── gcc
├── matlab
└── python_ast
```
or perhaps
```
cast
├── fortran
├── gcc
├── matlab
└── python_ast
```

 @vincentraymond-ua @titomeister @cl4yton 
Thoughts?

vincentraymond-ua:
@jastier - It looks like you've moved CAST2FN to the fortran directory instead of TS2CAST. CAST2FN is the CAST->Gromet pipeline.

cl4yton:
Re. naming: Good point, @myedibleenso 
I support moving to lower-case. However, let's keep references to CAST and FN (for Function Network) capitalized, so as to distinguish from "cast" (as in, conversion) and "function", respectively.

@vincentraymond-ua and @titomeister, what do you two think of this?

vincentraymond-ua:
@cl4yton - I also support keeping references to CAST and FN capitalized.

jastier:
@vincentraymond-ua 

Good catch.

Closing out this PR, will re-submit with the correct directory moved.

PR #375, opened 2023-07-28
Hackathon2 and M9 Report PR

titomeister:
This pull request contains all the changes done at the July 2023 Hackathon, and also contains fixes to ingest as much of the climlab model as possible. 
New additions
- Refactors the Operator handling so that the binary operator and the unary operator are done by two separate handlers, as opposed to having one giant if/else that checks the number of operators and acts accordingly.
- Added support for inserting source code files into the metadata, so it can be used for query lookup.
- Added unit tests for while1-while3 and cond1-cond2 examples. 
- Added helper function for determining the source of variables (local variable, function argument, global variable).
- Added support in the first three Annotated CAST passes to better support nested assignment. 

Fixes
- Fixes missing wiring (wfc, wfl, wopio) for conditional and loop examples by re-adding the logic for creating wires for loops/conditionals.
- Fixed wopio,wfc,wfl generation so that more than one can be created in the correct circumstances.
- Fixed overwriting variables issue in GroMEt FNs when assignment with function calls and attributes is involved.
- Fixed unpacking issue so that we can pack multiple return values into a single value that then gets returned. 
- Fixed a good deal of miscellaneous checks to make the GroMEt generation more robust.


vincentraymond-ua:
@titomeister - Could you pull master into your branch and handle the merge conflicts? It looks like there are some issues in multi_file_ingester.py. Other than that though it looks good to me!

titomeister:
> @titomeister - Could you pull master into your branch and handle the merge conflicts? It looks like there are some issues in multi_file_ingester.py. Other than that though it looks good to me!

Thanks Vincent! I've fixed the merge conflict now.

vincentraymond-ua (approved):

PR #377, opened 2023-07-28
Refactor TS2CAST to CAST/fortran

jastier:
Should be complete except for the class called TS2CAST, which I did not rename.


vincentraymond-ua (approved):
Thanks @jastier - Looks like everything is working on the Fortran side of things!

titomeister (approved):
The changes look good to me!

PR #385, opened 2023-08-03
Improved Error Handling mathml/amr, mathml/petrinet and mathml/ast-graph endpoints

Adi-UA:
#### Errors generated at endpoints
  - `/mathml/amr` and `/mathml/petrinet`. The following errors have been added
    - MISSING STARTING <math> TAG.
    - INVALID DERIVATIVE ON LHS.
    - MISSING EQUALS SIGN.
    - COULD NOT PARSE RHS.
    - MISSING ENDING </math> tag.
  - `/mathml/ast-graph`. The errors are only the generic `ParseError` generated by nom with no additional specificity. 
    -  Attempts to add specific errors to the generic MathML parser resulted in the errors only catching missing ending math tags even when it was only an intermediate tag that was missing. Adding specificity to `generic_mathml.rs` will require a deeper dive to solve.

#### Updates to ParseError implementation
 - Added a public function called `append_message` to the `ParseError` struct to allow for updating the error message as it is propagated upwards through nom and our own parsers. It only appends messages and never erases the existing error history in the object.
 - Updating the message in the ParseError requires nom's internal map function. To simplify this step of updating, the code for appending has been updated to a new public macro called `append_msg_to_parse_err`

#### Updates to `FromStr` trait implementation for first-order ODE and generic MathML parsers
- Previously, the `FromStr` trait was being used to parse MathML and FirstOrder ODEs from strings at the endpoints; however, now we use the `first_order_ode` and `math` functions. This was done because the `FromStr` trait doesn't allow lifetime declarations and can only pass back generic rust `Err` results containing a String message. The actual ParseError from nom cannot be propagated, which reduces flexibility for downstream applications.
 - The `FromStr` implementations previously did not propagate any errors and panicked instead, which halted program execution. Even though they cannot propagate ParseErrors, they have been updated to pass back regular Errs on failure so that any future users can catch and manipulate the errors as needed.

### Testing
- Passing `cargo test` and `cargo clippy`

### DIFF
--- a/skema/skema-rs/mathml/src/parsers/first_order_ode.rs
+++ b/skema/skema-rs/mathml/src/parsers/first_order_ode.rs
@@ -7,7 +7,9 @@ use crate::{
         Ci, MathExpression, Type,
     },
     parsers::{
-        generic_mathml::{attribute, equals, etag, stag, ws, IResult, Span},
+        generic_mathml::{
+            append_msg_to_parse_err, attribute, equals, etag, stag, ws, IResult, Span,
+        },
         interpreted_mathml::{
             ci_univariate_func, ci_unknown, first_order_derivative_leibniz_notation,
             math_expression, newtonian_derivative, operator,
@@ -17,6 +19,7 @@ use crate::{
 };
 
 use derive_new::new;
+
 use nom::{
     branch::alt,
     bytes::complete::tag,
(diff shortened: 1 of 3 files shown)


Free-Quarks (approved):
LGTM!

adarshp (approved):

PR #393, opened 2023-08-14
Lieffers hackathon cleanup

Free-Quarks:
This PR includes:

- Fixes bug with order of arguments for certain operators was getting messed up in parsing the AMR from the code
- Fixed some documentation in the code2amr endpoint (now gives code 200 if running example)
- Fixed some non-spec serialization for the AMR data struct (some None's were being serialized)
- Implemented parsing from FirstOrderODE for RegNet's
- Numerous clean ups and refactors of the AMR parsing code


adarshp (changes requested):
@Free-Quarks - can you fix the clippy warnings? I've attached the clippy warnings I saw here (see `clippy.txt`)

[clippy.txt](https://github.com/ml4ai/skema/files/12341094/clippy.txt)

Free-Quarks:
> 

I made some of the changes, but the rest (the majority) will have to wait until I do a proper refactoring of database.rs. Some of the changes they are suggesting would take me a bit to understand what is going on and if they are even applicable. The rest are mostly complaints about how I iterate over objects, and some depreciated extra arguments for functions, which is stuff I'll look for when I do that refactor. I would like to have that be the next PR, since we are building the code2amr integration testing infrastructure with TA4 and this gives proper and better amr's out which would be a good baseline.

I added some clippy allow conditions for things I think are easier to read too, even if it is slightly less efficient, such as if else statements with the same body. I think the consistent format of each condition and body, even if 2 are the same, looks better than adding a && statement into the mix, but this is a personal preference. 

There should be no compilation warnings related to my work now too, just clippy format warnings. (compilation warnings for expression.rs and comment_extraction still exist)

Hope that is okay!

adarshp:
Sounds good @Free-Quarks , thanks! I'll merge it now.

adarshp (approved):

PR #398, opened 2023-08-18
[tree-sitter] - Adds tree-sitter build system and tree-sitter comment extractor 

vincentraymond-ua:
## New tree-sitter build system

- Creates a new system for building and using tree-sitter parsers under `skema.program_analysis.tree_sitter_parsers`. 
- Updates Dockerfile-skema_py and existing CAST/fortran implementation to use new build system
- Removes tree-sitter-fortran submodule 
- Adds CI tests for building parsers
### Adding new parsers
See the [wiki](https://github.com/ml4ai/skema/wiki/Adding-support-for-a-new-tree%E2%80%90sitter-frontend) for information on how to add support for new languages.

### Building parsers
To build the parsers, run build_parsers.py with command line arguments representing the languages you want to build. 
```bash
python skema/program_analysis/tree_sitter_parsers/build_parsers.py --c --cpp --fortran --python
```

### Accessing parsers
You can access the shared object file by importing INSTALLED_LANGUAGES_FILEPATH like:
```python
from skema.program_analysis.tree_sitter_parsers.build_parsers import INSTALLED_LANGUAGES_FILEPATH
```

## tree-sitter comment extractor

- Adds a new comment extractor based off of tree-sitter under `skema.program_analysis.comment_extractor`.
- Creates endpoints for extracting comments and  checking language support
- Adds CI tests for extracting comments and accessing endpoints.
### Supporting new languages
To add a new language to the comment_extractor, add an entry to queries.yaml containing the queries to extract the different comment types. This assumes that you have already cloned and built the tree-sitter parser using the method above.

### Running the comment extraction service
To use the comment extraction service, run the following from the `skema/program_analysis/comment_extractor/` directory:
```bash
uvicorn server:app --reload
```
### Endpoints

- /comments-get-supported-languages: Returns the languages that are supported for comment extraction as well as the types of comments that can be extracted
- /comments-get-supported-file-extensions: Returns a list of file extensions that are supported for comment extraction
- /comments-extract: Extract comments from a source file
- /comments-extract-zip: Extract comments from zip archive of source files 

NOTE: The Rust comment extractor is not removed in the PR, as it is still being used for comment unification.


skema/program_analysis/comment_extractor/tests/examples/comments.c
```diff
@@ -0,0 +1,19 @@
...
+    on multiple lines
+*/
+
+/* Docstring in function declaration*/
+int foo();
+
+int foo(){
+	/*
+        Docstring in function definition
+    */
```
titomeister, line 16: Does the spacing on Line 16 matter here for the docstring? Does it need to match the spacing on line 14 or no? 
Ditto on the c++ example.

skema/program_analysis/comment_extractor/server.py
```diff
@@ -0,0 +1,86 @@
...
+import skema.program_analysis.comment_extractor.comment_extractor as comment_service
+from skema.program_analysis.comment_extractor.model import (
+    SingleFileCommentRequest,
+    SingleFileCommentResponse,
+    MultiFileCommentRequest,
+    MultiFileCommentResponse,
+    SupportedLanguageResponse,
+)
+
+SUPPORTED_FILE_EXTENSIONS = [".c", ".cpp", ".py", ".F", ".f", ".f95", ".for", ".m", ".r"]
```
titomeister, line 16: This is a super minor thing but would it make sense to maintain this SUPPORTED_FILE_EXTENSIONS list in terms of the EXTENSION_TO_LANGUAGE dictionary? Perhaps by computing the keys of the dictionary? It's a super minor thing but it would mean one less structure to manually update.

jastier (approved):
Looks good from here

titomeister (approved):
Aside from the very small comments that I left it all looks good to me! (Even the comments I left were just minor suggestions/questions and nothing too critical).

skema/program_analysis/comment_extractor/tests/examples/comments.c
vincentraymond-ua: The tree-sitter node structure is the same either way, so it wouldn't matter for extraction. However, I'll correct it just so the spacing is consistent.

skema/program_analysis/comment_extractor/server.py
vincentraymond-ua: I agree, that was kind of a last minute addition. I'm planning on updating that during the next sprint when we add support for comment unification.

PR #414, opened 2023-08-24
Reduce reader footprint

kwalcock:


enoriega (approved):
Looks great. Thanks for being thorough about it

enoriega:
@kwalcock We should reduce the processor instances to one soon. I am working on a benchmark extraction set to do unit tests. We can use it as a baseline for the current state and then open a new PR for further improvement

myedibleenso:
Great work, @kwalcock !  This is a huge cut in RAM usage.

Issue #388, opened 2023-08-07
The index page of the textreader webapp doesn't seem to work


kwalcock:
Submitting it just returns one to the home page.

kwalcock:
The problem seems to be that the brat/lib directory disappeared with [this commit](https://github.com/ml4ai/skema/commit/fe5f565865fa11cc2bffed55ad7cca7c16d3d221).  If I put it back, the webapp home page seems to run.  @myedibleenso, do you know what happened?

myedibleenso:
Yikes. I'm not sure why that would've ended up removed with the directory rename.  Does it have something to do with a .gitignore taking effect?

In any case, please add the directory back if it's needed.  I apologize for having removed it inadvertently. 🙇‍♂️

enoriega:
To be fair, we inherited a lot of code from automated that we don't use. I can see how this would've happen to me.

kwalcock:
This should be fixed with #415 as soon as it gets merged.  I think that a test will be difficult.  It is the web browser that can't find the files and load the JavaScript code.  We would have to find something that really loads the page and runs its code and then simulate a button press, AFAIK.

kwalcock:
I could just try to load the lib files by sending the right request to the server.  That would probably detect the absence of those particular files.

enoriega:
Fixed by #415

PR #415, opened 2023-08-24
Restore the brat lib files

kwalcock:
These files are needed so that the index page of the text reader works.  The PR will be separate from other reader changes that may soon appear.


kwalcock:
See also #388

myedibleenso (approved):
Thanks, @kwalcock , for fixing my mistake : )

It would be great to begin adding tests for the Scala REST API to avoid this sort of problem in the future.  Another PR?

enoriega (approved):

PR #418, opened 2023-08-25
[code2fn] - tree-sitter comment enrichment and unification

### LINKED ISSUE #266: [comment extraction] migrate to tree-sitter for comment extraction
With our adoption of tree-sitter, we want to revisit how we approach comment extraction to easily scale extraction to new languages.  

Comments are captured in tree-sitter grammars and tree-sitter queries can be used to extract them (ex. `(comment)+`).

```[tasklist]
### Tasks
- [x] Replicate comment extraction functionality in Python using `tree-sitter`
- [x] Remove comment extraction service from skema-rs
- [x] Add REST endpoint to skema-py for comment extraction
- [ ] Update e2e.ipynb example to use skema-py endpoint for comment extraction
```

### LINKED ISSUE #272: [comment extraction] Update single file endpoint to support multiple languages
Currently, the `extract-comments` endpoint in the skema-rs service assumes that the code being sent is Python. The `extract-comments-from-zipfile` endpoint can handle zipfiles containing a mix of Python, C/C++, and Fortran.

We may want to update the `extract-comments` endpoint to support multiple languages, depending on when this functionality is needed (e.g., if it is needed for this week's demo). 

If it is not needed urgently, then we should not update it and just go with the tree-sitter approach (#266).

### LINKED ISSUE #311: [skemy-py] Update fn-given-filepaths-zip to support comment unification

vincentraymond-ua:
## Automatic detection of supported file extensions
Adds extension information to languages.yaml in `skema/program_analysis/tree_sitter_parsers/` This allows us to automatically determine which extensions are supported for ingestion and create a mapping from extension->language, which was previously done manually.

## tree-sitter comment enrichment and unification
Replaces calls to the comment_proxy Rust service with calls to the tree-sitter comment extractor when doing system enrichment. Also rewires fn_unifier.py to accept tree-sitter comment response objects.

## Removes Rust comment extraction service
Removes the following Rust comment extraction components:

- comment_extractor Rust service
- comments_proxy endpoints

## Re-enables support for non-Python ingestion in fn-given-filepaths
Due to limitations with the Rust comment extraction endpoints, ingestion of non-Python source code would fail when using fn-given-filepaths and fn-given-filepaths-zip. This has now been resolved and enables the ingestion of any source files matching an extension returned by fn-supported-file-extensions

Resolves #266 
Resolves #311 
Resolves #272


myedibleenso (changes requested):
@vincentraymond-ua , great!  I think @adarshp should probably be added as a reviewer to make sure there isn't anything lingering in the Rust side of this. 

@vincentraymond-ua , I would like to see tests for all supported languages.

The Rust version had tests for C and C++.  Are those languages still supported?   If not, is that going to cause us any issues?

vincentraymond-ua:
@myedibleenso - I've added @adarshp as a reviewer

We do have a set of tests for the comment extractor that was added in the previous pull request. They are located here: https://github.com/ml4ai/skema/tree/main/skema/program_analysis/comment_extractor/tests

adarshp (approved):
LGTM!

titomeister (approved):
Looks good to me!

myedibleenso (approved):

PR #419, opened 2023-08-25
[fortran] Adds support for compound conditional

### LINKED ISSUE #379: [fortran] Support compound conditional idiom
A compound conditional is a conditional with multiple logical expressions. For example: 

```fortran
    if(x<2 .and. y>3) then
        ...
    end if 
```
Due to short circuiting, this is not compatible with out current Gromet schema. Our solution is to modify the code to convert this from a single conditional to a tree of conditionals. It was decided that the most convenient place to make this modification is at the CAST level rather than the Gromet level.

Update 8/2/2023 - I've implemented partial support for compound conditionals. However, more testing needs to be done for more advanced examples. Additionally, I'm currently only supporting compound conditionals in the top level if conditional. However, a compound could also occur in the elseif conditional.

vincentraymond-ua:
## Compound conditional
Adds support for compound conditionals to the Fortran->CAST frontend. Additionally fixes support for regular conditional when conditional statements are separated by comments.
 
## TIE-GCM fixes
Fixes the following bugs discovered when testing TIE-GCM
- Missing variable node when accessing an attribute of an imported derived type object.
- Missing operator node when using the '/=' operator

Resolves #379


titomeister (approved):
This looks good to me! Your handling of the compound conditional is way simpler than what I did in Python. When I get to doing the tree-sitter Python front end I'll have to take some notes from this.

PR #453, opened 2023-09-08
[TextReading] Removed Space Weather embeddings

enoriega:
This will help reduce the image size


kwalcock:
@enoriega, I was too slow.  Thanks!  AFAIK the grounder configuration is read in `TextReadingPipeline.newGrounder` which will find it in the Grounding section of `application.conf` set to Epidemiology.  It may be that the climatechange model can also be commented out.  Testing now...

enoriega:
I think we should have separate images for climate change and epi. That way we can cut more and still use them quickly on demand

kwalcock:
Some files/jars might be filtered out as the docker image is being built.  That way the model is still available for development.  It just doesn't get deployed.

PR #463, opened 2023-09-08
Unicode support for newtonian derivative

deepsana:
Incorporated Unicode support for Newtonian derivative:  

Unicode for `˙` is  `&#x02D9;` in MathML representation for Newtonian derivative.


Free-Quarks (approved):
LGTM!

PR #474, opened 2023-09-12
Updates to parsers

deepsana:
Changes in this PR:
- Struct `Ci` now has a field called `func_of` that takes in `Vec<Ci>` in which it handles the equations dependence explicitly. (e.g. , equation where we have variable `S(x,t)` now considers `x` and `t` as func_of). 
- Struct `Derivative` now has a field called `with_respect_to` that takes in `Ci` to keep track of what the derivative is taking with respect to is.
- First order derivatives now considers both `dS/dt` and `dS(t)/dt` as a function with either empty `func_of` field (without bounds) or time dependence `func_of` field (with bounds).
- The first order ODE parser now uses fields `func_of` and `with_respect_to` to keep track of what is in the first order ODE.
- `[` and `]` is now identified as left parenthesis and right parenthesis,respectively. 
- Msup can now interpret exponential and power as operators.
- Trig is now considered as an operator (e.g. cos(x) considers cos as an operator.)
- Content mathml now handles `with_repect_to` and `func_of` variables. 
    e.g. dS(t)/dt becomes          
       <apply><diff/>
              <bvar>t</bar>
              <apply><ci>S</ci></apply><ci>t</ci>
       </apply>


Free-Quarks (approved):
I added my changes to make it work with my AMR code and also threw in a small bug fix to img2amr integration. But looks good to me!

PR #478, opened 2023-09-14
Check for the brat assets

kwalcock:
This is to make sure we don't again accidentally remove the brat library files. The test isn't conducted the usual way with a FakeRequest. That doesn't seem to work and there was indication in forums that the assets simply aren't available during test. Indeed, favicon.ico cannot be routed, for example. So, I'm just checking that the files haven't been removed from the development environment.


myedibleenso (approved):
LGTM.  This is a step in the right direction.  Thanks, @kwalcock !

PR #479, opened 2023-09-14
[fortran] Support for fixed-form source and pinned tree-sitter-parser

### LINKED ISSUE #411: [fortan] Investigate and support Fixed-Form Fortran code
Currently we've been targeting Free-Form Fortran code and the Fixed-Form idioms that appear in the TIE-GCM model. We should investigate:

- Which Fixed-Form idioms is tree-sitter not able to parse?
- Can we use a tool like [fixed2free ](https://github.com/ylikx/fortran-legacy-tools) as a preprocessing step to convert Fixed-Form code to Free-Form automatically?
- If not fixed2free, can we add additional step(s) to the preprocessor to convert these idioms into a form supported by tree-sitter-fortran like how we've done with continuation lines?

vincentraymond-ua:
## Support for converting fixed-form Fortran to free-form
Attempts to convert fixed-form Fortran source code to free-form before parsing with tree-sitter. Since there is no consistent way of checking if a Fortran source if fixed-form, we run the tree-sitter parser both before and after converting to free-form to check if either returns a valid parse tree.

### Additional changes
- Removes existing implementation for handling fixed-form continuation lines.
- Adds testing framework for testing parsing TIE-GCM with tree-sitter.

## Support for pinning tree-sitter parser version
Adding an optional `commit_sha` field to languages.yaml to specify the commit to use when building a tree-sitter parser. If not specified, the most recent commit will be used instead. 

```YAML
fortran:
  tree_sitter_name: tree-sitter-fortran
  clone_url: https://github.com/stadelmanma/tree-sitter-fortran.git
  commit_sha: f73d473
  supports_comment_extraction: True
  supports_fn_extraction: True
  extensions:
    - .f
    - .F
    - .f90
    - .f95
    - .for
```

Resolves #411 
### Additional Changes
- Pins Fortran parser version to f73d473, enabling support for computed GO TO statements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,9 +114,7 @@ addopts = "--cov=skema --cov-report html:docs/coverage/python"
 testpaths = [
     "skema/program_analysis/tests",
     "skema/img2mml/tests",
-    "skema/rest/tests",
-    "skema/skema_py/tests",
-    "skema/program_analysis/tree_sitter_parsers/tests"
+    "skema/rest/tests"
 ]
 
 # Configuration for Black.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,7 +114,10 @@ addopts = "--cov=skema --cov-report html:docs/coverage/python"
 testpaths = [
     "skema/program_analysis/tests",
     "skema/img2mml/tests",
-    "skema/rest/tests"
+    "skema/rest/tests",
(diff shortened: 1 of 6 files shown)


myedibleenso:
@titomeister , do you think you can submit a review for this one today?

titomeister (approved):
All the changes look good to me!

PR #480, opened 2023-09-15
Increased the timeouts to handle long PDFs gracefully

### LINKED ISSUE #476: [TextReading] Play! Web Service resets connection with client for long inputs
## Problem

The [cosmosJsonToMentions](https://github.com/ml4ai/skema/blob/df1f8066245452bf7313d6e78595b7399ca862df/skema/text_reading/scala/webapp/app/controllers/HomeController.scala#L86) endpoint on the Play! REST api is blocking and times out for long papers, such as the SIDARTHE. 

Upon debugging I discovered that the underlying server process handling the request closes the connection even if the controller method runs to completion after the request is left orphaned.

I attach the cosmos json input that replicates this issue
[sidarthe.zip](https://github.com/ml4ai/skema/files/12603213/sidarthe.zip)

enoriega:
Fixes #476

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -66,6 +66,7 @@ jobs:
         password: ${{ secrets.LUM_ASKEM_DOCKERHUB_TOKEN }}
     - name: "Free disk space"
       run: |
+        sudo apt update
         sudo apt clean
         sudo rm -rf /usr/share/dotnet
         sudo rm -rf /opt/ghc
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -66,7 +66,7 @@ jobs:
         password: ${{ secrets.LUM_ASKEM_DOCKERHUB_TOKEN }}
     - name: "Free disk space"
       run: |
-        sudo apt update
+        sudo apt-get update
         sudo apt clean
         sudo rm -rf /usr/share/dotnet
         sudo rm -rf /opt/ghc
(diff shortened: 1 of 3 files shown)


enoriega:
@kwalcock Tested the docker image built locally and works!

kwalcock:
It looks very good to me.  I read some of the assertions about the timeout not being respected in dev mode, and it looks like you found the solution.  Thanks for getting to it before I did.

Does 40 minutes processing time indicate some other problem?

The CI has been failing on the docker container for me as well.

kwalcock (approved):

enoriega:
@kwalcock The 40 mins is just some arbitrary value, long enough to process a paper with many dozens of pages and accounting for COSMOS processing time.

CI is failing for something apparently unrelated. Not sure yet how to fix it

enoriega:
@myedibleenso @kwalcock It seems that there is an issue in the `free disk (aggressively)` action of the GitHub workflow where it can't access an ubuntu repository correctly. I tried a couple variants of a fix I found online, but didn't succeed on fixing it. Do you have any leads on how to fix it?

myedibleenso:
I think the CI issue is the same one reported here: https://github.com/jlumbroso/free-disk-space/issues/9

It looks like the repo that hosts the action has the same failing CI check after a recent PR was merged.

I'm thinking you could try setting this line to `false` and tossing a pinch of salt over your left shoulder: https://github.com/ml4ai/skema/blob/ceb20ee20ff309c63e3115adf5e50a53e4a58365/.github/workflows/docker.yml#L266

Not sure if this will work, as I wouldn't expect a recent change to that action to affect a previously tagged version of the action...

enoriega:
@myedibleenso Testing after applying the fix

PR #482, opened 2023-09-15
[CI] Use fork with fix to jlumbroso/free-disk-space/issues/9

myedibleenso:
## Summary of changes
Resolves a CI issue related to an action that removes unused packages to create additional space on the runner for docker builds

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -257,7 +257,8 @@ jobs:
     # The TR image is huge.
     # We need to free up every last bit of space we can.
     - name: Free disk space (aggressively)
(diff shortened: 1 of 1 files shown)


myedibleenso:
FYI @enoriega @kwalcock

enoriega (approved):
LGTM

kwalcock:
Thanks for tracking that down.  Does there need to be an update after the PR is merged into the official version?

myedibleenso:
@kwalcock , we might want to update it then.  Alternatively, we might just want to fork it (perhaps to `lumai`?) and use that version.  It shouldn't need to change very much and hopefully we can phase it out completely after reducing the image size for TR.

PR #483, opened 2023-09-15
GroMEt Function Call Handler Rewrite PR

### LINKED ISSUE #197: [python] Bug in ingesting python lotka-volterra implementation
Currently Python2CAST breaks when ingesting this file:
https://github.com/smkalami/lotka-volterra-in-python/blob/b126ef16c1c072847dc5fbc9c3afaa549e6d49b1/app.py#L42-L43

- for-loop on lines 42-43
- looks like it might be b/c they have a user-define function 'next' that may be clashing.

### LINKED ISSUE #380: Missing Port Boxes in EvaluationScenario3
Justin pointed out a bug that some ports are missing the boxes they belong to.
A specific example of this can be observed in the EvaluationScenario3 GroMEt FN in line 463.

Investigate what's causing this and fix it.

### LINKED ISSUE #406: [Gromet] Refactor of GroMEt Function Call Handler
The function call handler for GroMEt generation has gotten very difficult to maintain, and needs some serious refactoring. 
Rewrite the handler for function calls so that it's more consistent and easier to maintain. 

```[tasklist]
### Tasks
- [ ] Rewrite the AnnCastCall handler in to_gromet_pass.py
- [ ] #380 
- [ ] #197 
```

### LINKED ISSUE #434: [gromet] - Python multiline comments converted to List LiteralValue
Python multiline comments are being converted to a List Literal Value in Gromet. The list contains a single string representing the comment. This isn't a major issue, but is causing confusion in the execution engine. I think since comments are now being included in Metadata, we should remove this.

For an example, see the following Gist:
https://gist.github.com/vincentraymond-ua/095e50f5e2217e52c6d863a2bc149916

titomeister:
### Refactoring of GroMEt Function Call Handler
- The handler that creates the GroMEt elements related to function calls needed a rewrite due to it becoming very difficult to maintain and change when bugs were discovered.
- The main goal was to make the code significantly smaller, and to remove redundancy and streamline the process as much as possible.
- The biggest redundancy changes was with the argument handling, as it was being done an unnecessary amount of times.
- There was also inconsistency with pof port generation and then it was needed and not needed. This caused inconsistent wiring generation.  
- A test file was added to support checking Gromets of small example programs that exercise the generation of function calls for both normal and primitive function calls.

### Removing Docstrings from Python AST to CAST
- Python stores multi-line comments as docstrings. These don't get eliminated from the AST, as they get stored as normal strings.
- This causes some issues when creating the CAST, because they are left in there as floating strings that don't contribute much to the extraction from this perspective. They're more relevant when they're extracted as comments.
- I've gone ahead and added code that removes all instances of docstrings that appear at the function definition level, global level, and class definition level. Now there'll no longer be any floating docstrings in the CAST. 
- The only strings that remain in the CAST are ones that are used as an actual value in a computation (assignment, function call, etc..)

### Additional Changes
- Removing commented out lines of code that aren't needed.
- Minor formatting fixes done with Black autoformatter.

Resolves #406 
Resolves #380 
Resolves #197 
Resolves #434

### DIFF
--- a/skema/program_analysis/CAST/pythonAST/py_ast_to_cast.py
+++ b/skema/program_analysis/CAST/pythonAST/py_ast_to_cast.py
@@ -86,6 +86,23 @@ def construct_unique_name(attr_name, var_name):
     """
     return f"{attr_name}.{var_name}"
 
+def check_expr_str(ast_node):
+    """
+        Checks if this node is an Expr node
+        with a string value
+        This is added to remove multi-line comments
+        that otherwise don't do anything
+
+        Current cases for multi-line comments
+        - FunctionDef
+        - Module level
+        - ClassDef level
+    """
+
+    if isinstance(ast_node, ast.Expr):
+        return isinstance(ast_node.value, ast.Constant) and isinstance(ast_node.value.value, str)
+    return False
+
 
 def get_node_name(ast_node):
     if isinstance(ast_node, ast.Assign):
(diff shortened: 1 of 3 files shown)


vincentraymond-ua (approved):
@titomeister - Looks good!

PR #484, opened 2023-09-18
Testing skema/skema_py/tests

vincentraymond-ua:


myedibleenso:
@vincentraymond-ua , you may want to checkout `.github/workflows/tests-and-docs.yml` from `main`.

PR #487, opened 2023-09-18
Use skema[ml] as dep in skema[metal]

myedibleenso:
## Summary of changes

Adds `skema[ml]` as a dependency in `skema[metal]` to more easily enforce dependency versions of PyTorch and related deps.


enoriega (approved):
LGTM

PR #502, opened 2023-09-22
[morae] Update to Memgraph LiteralValue representation

### LINKED ISSUE #402: [MORAE] - Improve representation of Literal Value
Currently :Literal nodes are represented like:
`"value: "ValueL { value_type: "Integer", value: "10", gromet_type: Some("LiteralValue") }"`

This requires extra string manipulation to parse out the value and value_type. Could we instead have Value and Value type as top level fields on the Node?

vincentraymond-ua:
## Updated representation of LiteralValue
Previously LiteralValue objects would be serialized into a single long debug string before being uploaded to Memgraph. For example: 

`"value: "ValueL { value_type: "Integer", value: "10", gromet_type: Some("LiteralValue") }"`

This has been updated to upload the LiteralValue as an object instead, allowing easier access to each individual field.

269936255-c56289bf-077e-437c-9aa6-d055014e01a7

Resolves #402


Free-Quarks (approved):
Yeah this looks good to me!

PR #503, opened 2023-09-25
database.rs refactor #1

Free-Quarks:
This is an initial database.rs refactor.

It should have cleaned up the code on a high level some.
Also should have fixed some wiring bugs Vincent had mentioned.
Also should not crash on taking in while loops or conditionals now.

This is the first PR of 2 this refactor has been split into, so as to not be a blocker for execution engine development. 

Things for the next PR 2:
- Fix wiring bugs for while loops and conditionals
- add support for For loops
- clean code on cargo clippy level
- Fix bugs in nested function wiring I expect, but haven't tested yet


adarshp (approved):

adarshp:
This does not depend on the TR pipeline, so I am merging it.

PR #505, opened 2023-09-26
GroMEt and Metadata Version Updates

### LINKED ISSUE #506: GroMEt and Metadata Version Update
The GroMEt and Metadata versions are out of sync. We need to update the GroMEt version so it matches the metadata version.

We need to create a new Metadata type that we can use to send debugging messages back to the user whenever something goes wrong in any of the endpoints that return GroMEt. 

(NOTE: At the time of writing this issue the problem has already been resolved. I'm creating the issue mostly to link the PR associated with this so we can track it in the issue board)

titomeister:
This PR updates the version of GroMEt and Metadata to version 0.1.8. This means that GroMEt and Metadata have the same version again, and will continue to have the same version from now on.

### GroMEt Changes
- None. This is strictly a version update to sync up with the metadata version.

### Metadata Changes
- Adds a 'Debug' metadata type which can be used to return warning and error messages back to the user if something is wrong with their input. The current use case is to use the Debug metadata to give a message back to the user if any of the files they submitted for processing are not supported.
- The script `codegen_swagger_models.py` now also updates a variable in the file `skema/gromet/__init__.py` that maintains the current GroMEt version number, so that we don't have to manually update it.

Resolves #506


scripts/codegen_swagger_models.py
```diff
@@ -34,15 +35,31 @@
...
+
+
+# -----------------------------------------------------------------------------
+# Update Canonical GROMET_VERSION
+# -----------------------------------------------------------------------------
+
+def update_canonical_gromet_version(new_version):
+    path = os.path.join('..', GROMET_VERSION_CANONICAL_DEFINITION)
+    with open(path, 'r') as gv:
+        file_contents = re.sub(r"\s*GROMET_VERSION\s*=\s+\"[\d.]+\"",
```
myedibleenso, line 57: I think we should change the regex to `r"\s*GROMET_VERSION\s*=\s+\".*?$"`

myedibleenso (changes requested):

scripts/codegen_swagger_models.py
titomeister: Done!

myedibleenso (approved):
LGTM.  Thanks, @titomeister !  Let's wait to merge until the TR service is back up and we can re-run our CI tests.

vincentraymond-ua (approved):
Looks good to me too!

PR #507, opened 2023-09-26
[Text Reading] Set `_JAVA_OPTIONS` to min. RAM requirements

myedibleenso:
## Summary of changes
Sets `_JAVA_OPTIONS` env variable to a default value that illustrates the minimum RAM requirements for the app.

This can be overwritten using the `-e` flag when launching a container using our TR image.


kwalcock (approved):
LGTM.  Hopefully the file.encoding is not necessary.

PR #517, opened 2023-09-27
Incorporated multiply in MathML and fixed minor bugs

deepsana:
Incorporated being able to handle multiplication when we have <mo>*</mo> and <mo>& #x2217;</mo>. Fixed minor bugs.


Free-Quarks (approved):
LGTM!

PR #520, opened 2023-09-27
[python] add missing deps

myedibleenso:
## Summary of changes

- Adds missing required deps (previously pulled in as transitive dependencies)
- Add missing optional deps as extras (ex. matplotlib)
- Removes unused file


titomeister (approved):
This looks good to me! It's eliminated all instances of the ast viewer, which we don't need anymore.

Issue #400, opened 2023-08-18
[MORAE] - Wiring bugs in Memgraph


vincentraymond-ua:
There a couple of wiring bugs when converting from Gromet JSON to the Memgraph graph representation. 

## Missing primitive WFOPI
When a primitive operator has an opi as an input (x=1+y) the wire from the primitive to the opi is missing.

## Missing function input wires
When user defined functions are called, their inputs are not correctly wired to the opi. 

## Missing index field on wires causing ambiguity
There are some cases where wires don't have index fields, but should have them to avoid ambiguity. In the following example there are 4 ast.Div :Primitive nodes that are all children of a top level :Expression node. Since there is not an index field, we don't know which primitive to start with. 
```Python
output = a/b/c/d/e
```

## Builtin functions not showing up in memgraph
Calls to builtin functions like `print()` and `min()` are not showing up in the memgraph database.

## Mysterious user defined function bug
There is a werid bug that causes the memgraph import to crash if a user defined function takes no input and is called more than once. I've confirmed that both of these things have to be true for the crash to occur. 
```Python
def foo():
    return 1+2
out = foo()
out_2 = foo()
```

vincentraymond-ua:
## Update 9/5/2023
Added additional wiring bug

vincentraymond-ua:
## Update 9/8/2023
Updated issue regarding missing function input wires to generalize it to all user defined function calls.

Free-Quarks:
@vincentraymond-ua I am going through some older issues that are still open and closing some of them out. 

I think I have fixed these bugs for now. At least to my knowledge. Does that seem correct to you too?

Free-Quarks:
I think this was fixed at some point, not sure when though

PR #521, opened 2023-09-28
[execution_engine] Initial support for parameter extraction and Gromet execution

vincentraymond-ua:
## Initial support for parameter extraction
Adds initial support for parameter extraction and Gromet execution. We are currently able to execute all builtin operators using a handful of common builtin data types. We can then keep track of each symbol and return either: the initial value in the case of parameter extraction or a history of that value over time.

### Example execution
```Bash
python execution_engine.py variable_test --host localhost --port 7687
```

**Input**
```Python
#variable_test.py
a = 10
b = a+30
c = b-a
```
**Output**
```
{'a': tensor(10, dtype=torch.int32), 'b': tensor(40, dtype=torch.int32), 'c': tensor(30, dtype=torch.int32)}
```

### LiteralValue Support
Extends support for LiteralValue types after the memgraph representation was updated in  #502.  

The execution engine supports representing basic data types. For more complex data types like List and Map, the values have to be decoded before they can be converted to Tensors.
```
[Supported]: Integer (i.e. 1)
[Supported]: Float (i.e. 2.1)
[Supported]: Boolean (i.e. True)
[Unsupported]: List (i.e. [1,2,3])
[Unsupported]: Map (i.e. {"key": "value"})
```

### Primitive Support
Extends support for Primitive operator execution after variable wiring bugs were fixed in #503. We are now supporting primitive operators with variable inputs. (a+b)

The execution engine supports executing all builtin primitive operators from the operator module. However, some input types are not currently supported. NOTE: `and` and `or` are not considered builtin operators for Python because of short circuiting. 

``` Python
a = 1 + 1             # [Supported] - Integer LiteralValue inputs
b = 1 + 3.1           # [Supported] - Float LiteralValue inputs
c = not True          # [Supported] - Boolean LiteralValue inputs
d = a + b             # [Supported] - Variable inputs
e = foo(a, b) + 11    # [Unsupported] - Function Call inputs
f = [1,2,3] + [4,5,6] # [Unsupported] - List inputs
g = {} | {}           # [Unsupported] - Dict/Map inputs
```

### Function Call/Def support
Function calls are not currently supported due to a few wiring bugs referenced #400. However, I have implemented the framework for keeping track of symbols and values through scopes in symbol_table.py. This allows us to not only access the initial value of a variable, but also any value that the variable represents during the execution of the source code.


skema/gromet/execution_engine/execution_engine.py
```diff
@@ -0,0 +1,198 @@
+import yaml
+import argparse
+import asyncio
+from pathlib import Path
+from typing import Any, List, Dict
+
+import torch
+from gqlalchemy import Memgraph
```
myedibleenso, line 8: I think this is missing from pyproject.toml

myedibleenso (changes requested):

skema/gromet/execution_engine/execution_engine.py
vincentraymond-ua: Ah yes you are right, do you think it should be added as a [dev] dependency, or as something else?

myedibleenso: I was thinking just under our list of project dependencies (which is where `tree-sitter` is currently): https://github.com/ml4ai/skema/blob/d3bc7d19fd87309c84c12c0961d5a43977a2f963/pyproject.toml#L19

pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies=[
     "networkx==3.1",
     "PyYAML",
     "tree-sitter",
+    "gqlalchemy",
```
myedibleenso, line 20: any expectations on a min. version?

skema/gromet/execution_engine/execution_engine.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if value_type == "Integer":
+            return torch.tensor(int(value), dtype=torch.int)
+        elif value_type == "AbstractFloat":
+            return torch.tensor(float(value), dtype=torch.float)
+        elif value_type == "Boolean":
+            return torch.tensor(value == "True", dtype=torch.bool)
+        elif value_type == "List":
+            # TODO - Add support for List
+            pass
```
myedibleenso, line 149: should this raise and exception for the time being?

myedibleenso (commented):
Any thoughts on how we can test the ee during CI?  It sounds like we need to have a memgraph service running, right?

vincentraymond-ua:
@myedibleenso - Yes currently the execution engine assumes that we have a Memgraph instance running already. Does the service at [https://skema-rs.askem.lum.ai](https://skema-rs.askem.lum.ai ) run Memgraph?

myedibleenso:
skema-rs communicates with a memgraph service that is run separately.

For CI, we could do something like this: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers

In our case, we would mirror the service entry we use in the DockerVM repo.

vincentraymond-ua:
Yeah I think that would work, this could be a good issue for next sprint or the following to create some initial tests for the execution engine.

pyproject.toml
vincentraymond-ua: I'm not certain, I've tested with 1.4.0 and 1.5.0 and know they both work with our current system. To prevent breaking changes, we can pin it to 1.5.0 which is the most recent version.

skema/gromet/execution_engine/execution_engine.py
vincentraymond-ua: The reason I'm not current raising an exception here is because an unsupported type isn't necessarily catastrophic. Take for example the following:  
```Python
a = 10
b = 20
c = [1,2,3] 
d = a+b
```
Here we are still able to extract parameter values for a, b, and d. With the current implementation, c will just show up as a None value. 

However, I am open to input as well.

myedibleenso: Thanks for the explanation.  Perhaps we should at least produce a warning?

pyproject.toml
myedibleenso: If we know 1.4.0 and 1.5.0 both work, let's just use 1.4.0 as the min. working version.

skema/gromet/execution_engine/execution_engine.py
vincentraymond-ua: Yeah, I like that idea. I'll add some logging messages.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies=[
     "networkx==3.1",
     "PyYAML",
     "tree-sitter",
-    "gqlalchemy==1.5.0",
+    "gqlalchemy==1.4.0",
```
myedibleenso, line 20: Ah, sorry for the confusion. I meant why not `"gqlalchemy>=1.4.0"`

pyproject.toml
vincentraymond-ua: It looks like there is actually a dependency version conflict with networkx. 

```
skema[all] 0.1.0 depends on networkx==3.1
gqlalchemy 1.4.0 depends on networkx<3.0.0 and >=2.5.1
```

Do you know if the version can be lowered?

myedibleenso: Ah, sounds like the min. compatible version might actually be `1.5.0`.  If we put `>=` for the `gqlalchemy` version (ex. `"gqlachemy>=1.5.0"`), does everything install for your locally?

myedibleenso (changes requested):

pyproject.toml
vincentraymond-ua: The same thing happens with 1.5.0 unfortunately. My skema install is from before we started pinning dependencies, so I have an earlier version of networkx.

myedibleenso (approved):

PR #522, opened 2023-09-28
Incorrect Metadata Fix 

### LINKED ISSUE #490: Incorrect Metadata Stored in Gromet Boxes
Vincent discovered in Gromet that certain box metadata is being assigned the actual Metadata object as opposed to indices into the Metadata table. 
This appears to be happening with SourceCodeReferences, and it appears to do with a missing line in the handle_primitive_function() method of the Gromet generation pass.

titomeister:
### Incorrect Metadata Type Fix
- The rewritten call handler was not properly inserting metadata into the metadata table, causing it to assign the metadata itself instead of the index into the metadata table (line 2665-2466 in `to_gromet_pass.py`). This has been fixed, and now the correct metadata index is being assigned.

### Other Changes/Fixes
- Code2FN now uses the `GROMET_VERSION` variable located in `skema/gromet/__init__.py` for GrometCreation metadata and schema_version in the GroMEt JSON.
- Changed the import types in the FNs from "Expression" to "Import".
- Minor cleanups

Resolves #490


vincentraymond-ua (approved):
@titomeister - Looks good, thanks for making this fix!

PR #535, opened 2023-09-29
continue on error for TR docker builds

myedibleenso:
## Summary of changes

The CI step for building the TR docker image often fails or takes an inordinate amount of time to build and publish (ex. 2 hours).

Culprits include the the clulab JFrog repo which sometimes times out when we attempt to download (big) artifacts and the issues related to the large image size.

Our AWS deployment does not include the TR service (but may in the future if RAM consumption is reduced).  Until this PR, a failing TR image build & publish step would prevent a redeployment of our AWS service.


enoriega (approved):
LGTM

PR #538, opened 2023-10-02
[CI] Make TR docker build its own workflow

myedibleenso:
## Summary of changes

- Make TR image creation and publishing its own workflow
- Update changes.md with notes on pending release


titomeister (approved):
Looks good!

PR #542, opened 2023-10-03
[code2fn]  Adds debug messages when ingesting unsupported files

### LINKED ISSUE #401: [code2fn] - What happens when passing an unsupported file to the code2fn endpoints?
We want to make sure we are properly handling the case where an unsupported file extension is passed to the code2fn endpoints.

vincentraymond-ua:
## Debug messages when trying to ingest unsupported files
Adds debug messages when the user tries to ingest an unsupported file type through the code2fn endpoints. The messages are passed through the Gromet as a Debug Metadata type in the top level `GrometFNModuleCollection.metadata_collection` field.

### Debug message types
- WARNING: In the case there are mixed supported/unsupported files in a system.
- ERROR: In the case there are are only unsupported files in a system.
### Testing
- Adds two additional tests, `test_no_supported_files` and `test_mixed_supported_files`
- Re-enables skema_py endpoint tests

## Updates to Gromet FN schema
Adds a metadata_collection field to GrometFNModuleCollection. This change is still considered part of Gromet FN 0.1.8, so the version number will not be bumped.

Resolves #401


titomeister (approved):
Looks good to me!

PR #545, opened 2023-10-03
Python tree-sitter to CAST porting: Arithmetic Operations

### LINKED ISSUE #523: Python tree sitter: Arithmetic

- Getting end-to-end (Source to GroMEt) generation for arithmetic operations

titomeister:
This PR introduces support for generating CAST for Arithmetic Operations using tree-sitter, as part of the ongoing effort to port over the Python AST to CAST generation to using tree-sitter.

### Python Tree Sitter

- Adds additional handlers to support arithmetic binary operators (+,-,/,*) and unary operators in tree-sitter. Also adds handling for parenthesized operations (which is currently a simple pass-through).
- Added some more support code for handling operators.
- Added Code2FN support for the arithmetic operators.

### Testing

- Added a pytest script to test the CAST structure of Arithmetic Operators.
- As of now the test script tests simple CAST features of the Operators, but as more features get ported over the scripts could get more complex.

Resolves #523


skema/program_analysis/CAST/python/ts2cast.py
```diff
@@ -148,7 +156,29 @@ def visit_assignment(self, node: Node) -> Assignment:
...
             source_refs=[literal_source_ref]
         )
+
+    def visit_unary_op(self, node: Node) -> Operator:
+        """
+            Unary Ops
+            OP operand
+            where operand is some kind of expression
+        """
+        op = self.node_helper.get_operator(node.children[0])
```
vincentraymond-ua, line 166: It likely doesn't matter in this case, but be careful when accessing the children of a node by index. There are cases where the structure may change.

For example, a comment node can shift all of the children indexes by 1. 

This was one of the motivations for the various helper functions in node_helper.py.

vincentraymond-ua (approved):
@titomeister - Looks good!

skema/program_analysis/CAST/python/ts2cast.py
titomeister: Oh, that's super important to know actually. That makes a lot of sense. I think as part of the next feature I port over I'll add helper functions like you have in order to defend against this, because that will most definitely come up somewhere. Thank you for that!

PR #560, opened 2023-10-10
MATLAB Tree-sitter binary operator support

jastier:
## MATLAB Tree-sitter binary operator support
This PR adds CAST parser and testing support for the `binary_operator` node of the MATLAB grammar.  

 ```matlab 
z = x + y;
z = x + y / 4;
... 
```
## Relavant features
- Operations can be chained together
- CAST structure follows PEMDAS rules
- A test case for a MATLAB code snippet with a binary_operator node is provided
- GroMEt output is produced but not yet tested

## Related issues
- Resolves issue #532
- Partially resolves issue #420

### CODE REVIEW

**Comment by myedibleenso** [comment]
@titomeister , can you please review this one?

File: skema/program_analysis/CAST/matlab/tests/test_assignment.py [hunk 1]
```diff
@@ -37,3 +37,6 @@ def test_single_assignment():
     assert module['left']['val']['name'] == 'x'
     assert module['right']['value_type'] == 'Integer'
     assert module['right']['value'] == '5'
+
+if __name__ == "__main__":
```
titomeister, line 41: This is more of a curiosity, but do you have to have these 'if main' checks for the tests to run on your end? 
When I write my test scripts I don't have to put these in, since pytest just collects every function that starts with a 'test' and runs that way. When I want to run them myself I just do 'pytest FILE' and that executes any functions that start with test in FILE (whatever FILE is).

skema/program_analysis/CAST/matlab/tests/test_binary_operation.py
```diff
@@ -0,0 +1,50 @@
...
+    assert len(nodes) == 1
+
+    body = nodes[0]['body']
+
+    # There should only be one module within the node body 
+    assert len(body) == 1
+
+    # Test the module dictionary for relevant key values
+    module = body[0]
+    assert module['node_type'] == 'Assignment'
```
titomeister, line 39: I think this is a fine way to access the CAST, but whenever you get a chance you should look through my test scripts to see how I access the CAST using the objects instead of the dictionary. 
Here's an example [of a test script for expressions](https://github.com/ml4ai/skema/blob/main/skema/program_analysis/tests/test_expression_cast.py)

Honestly the dictionary way might be cleaner, but I'm used to doing it with the objects.

titomeister (approved):
Looks good! I left some comments as suggestions, I recommend looking at those.

jastier:
Great ideas Tito!  I can do it that way.

PR #567, opened 2023-10-12
[fortran] Bug fixes for ingestion TIE-GCM source

### LINKED ISSUE #563: [fortran] Fix bugs in TIE-GCM ingestion
There are a few files that are generating CAST, but cause an error during Gromet ingestion, implying that the CAST is malformed. This occurs in the following TIE-GCM files:

- advec.F
- cons.F
- colath.F
- qjnno.F

vincentraymond-ua:
## TIE-GCM bug fixes
There were a number of small bugs in the Fortran->Gromet pipeline that were discovered when determining which TIE-GCM files are currently able to be ingested. 

### Node Helper
- Fixes missing comma in CONTROL_CHARACTERS list causing '>' to be ignored. 
- Adds additional set of control characters to CONTROL_CHARACTERS list that weren't previously seen in any examples
- Adds remove_comments function to remove comment nodes from tree-sitter parse tree. This prevents comments from being processed by ts2cast and creating unexpected None values in the CAST.

### Preprocessor
- Fixes bug in include directory path when running script from directory other than the system root. 
- Runs gcc processor from system root if there are no `#include` directives. 

### TS2CAST
- Fixes bug in function call handler to support ingesting functions with no arguments. 
- Fixes bug when in function call handler when the function call is part of a unary expression `+foo()`
- Fixes bug in conditional handler to support ingesting single line conditionals
- Adds support for `not` operator

Resolves #563


jastier:
Nothing jumps out as a potential problem, well commented, good edge case checking.

LGTM

jastier:
Nothing jumps out as a potential problem, well commented, good edge case checking.

LGTM

jastier (approved):
I don't see any issues.   Well commented, good edge case checking.

LGTM

PR #572, opened 2023-10-13
Added support for cosmos json files

enoriega:
We now support processing cosmos json files directly from the `integrated-text-reading` endpoint.
If a pdf is preprocessed, the intermediate artifact will be used directly, speeding up things

### DIFF
--- a/skema/rest/integrated_text_reading_proxy.py
+++ b/skema/rest/integrated_text_reading_proxy.py
@@ -470,7 +470,11 @@ async def integrated_pdf_extractions(
     # Call COSMOS on the pdfs
     cosmos_data = list()
     for pdf in pdfs:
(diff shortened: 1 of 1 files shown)


enoriega:
@brandonmr FYI

enoriega:
@brandomr fyi

brandomr:
@enoriega that's great--can you give an example of how we should provide the cosmos extractions over to you?

enoriega:
@brandomr Using the same format that our `cosmos_to_json` endpoint generates. I believe that endpoint is replicated in the knowledge middleware. I attach a sample file here for reference.

[response_1697603574951.json](https://github.com/ml4ai/skema/files/12982974/response_1697603574951.json)

PR #575, opened 2023-10-13
[Code2FN] Gromet FN Missing Function Types Fix

### LINKED ISSUE #537: [code2FN] Certain plotting boxes don't have a function_type field
I noticed some boxes related to plotting seem to be missing a function_type field in the FN. I believe this is a bug. 

the json: https://gist.github.com/Free-Quarks/e63b6aff1c61a90010cd18fb5d3453fe#file-sidarthe_full-json

the code: https://gist.github.com/Free-Quarks/e63b6aff1c61a90010cd18fb5d3453fe#file-sidarthe_full-py

### LINKED ISSUE #581: [code2FN] Empty Function boxes where it should be import_method
There seems to be some empty function boxes being generated instead of what I believe should be imported_methods. This seems to be related to calls when "import as" is used in the import. Such as np.linspace when numpy is imported as np.

titomeister:
This PR contains a couple of changes to the Gromet generation to fix certain Gromet FN call boxes not containing what kind of function type they were.

### Missing function_type Fix
This issue manifested when the function call was type 'Attribute' for a variable that contains an object which we don't have any real information about. In this case, the Function Type should be "IMPORTED_METHOD" with an import type of "OTHER".  Some additional code was added to handle this case appropriately.

### Missing import_method Fix
This issue manifested when a function call with type 'Attribute' for a Python library was done. As an example: calls from the numpy library were missing the function_type and the import_type. Some additional handling was added to support this. 

Resolves #537 
Resolves #581


vincentraymond-ua (approved):
@titomeister - Thanks Tito, looks good to me!

Issue #570, opened 2023-10-12
CAST Node testing


jastier:
Refactor MATLAB tests to use CAST Node objects instead of python dictionaries.  
This will create tests similar to what is being done now.

jastier:
This issue will be resolved by PR #563 once it is merged.

PR #583, opened 2023-10-17
CAST based MATLAB testing

jastier:
## Testing of MATLAB idioms using CAST structures
This PR is a refactor of the original test plan where MATLAB parser output was tested as a python dictionary.   CAST testing uses fewer lines of code to achieve the same results and follows existing SKEMA test plans.

## Relevant Features
- CAST elements are used to test MATLAB idioms
- Source code is now defined within each test file
- The test file data directory is no longer needed and has been removed
- The MATLAB parser now takes file or string source code input

## Related Issues
- Completes issue #570


skema/program_analysis/CAST/matlab/tests/test_assignment.py
```diff
@@ -1,42 +1,35 @@
...
-
 def test_assignment():
-    """ Tests whether the parser can handle a single assignment statement """
-    filepath = Path(__file__).parent / "data" / "assignment.m"
-    print(f"Testing parse tree for {str(filepath)}")
-    parser = MatlabToCast(filepath)
-    cast = parser.out_cast
+    """ Tests parser assignment CAST """
+
+    source = 'x = 5;'
```
titomeister, line 14: I think hardcoding the source like this is okay to do for the tests right now, I do effectively this in my test scripts, but I want to note that at some point we will be changing tests to access test files that are online on a google drive. I think we just haven't decided how to go about doing that yet.

titomeister (approved):
Looks good to me!

skema/program_analysis/CAST/matlab/tests/test_assignment.py
jastier: Sure, that's how I had been doing it.   I can easily switch back if needed.

PR #584, opened 2023-10-17
Python tree-sitter to CAST Porting: Function Definition

### LINKED ISSUE #500: Python tree sitter: Functions

- Getting end-to-end (Source to GroMEt) generation for Function Definition and Call structures (For and While loops)

titomeister:
This PR introduces support for generating CAST for function definitions and other related idioms, using tree sitter, as part of the ongoing effort to port over the Python AST to CAST generation using tree-sitter.

### Python Tree Sitter
- Adds additional handlers to support generation for Function Definition, Call, ModelReturn CAST nodes.
- Brought over extra utility functions from the Fortran implementation for working with the tree-sitter nodes.
- Added extra handling for generating other LiteralValue types (Boolean, Floats).
- Began adding type hints to the visitors.

### Testing
- Added a pytest script to test the CAST structure for Function Definition, Call, and ModelReturn nodes.

### Miscellaneous Fixes/Changes
- Fixed a discrepancy for when a Var node and a Name node should be used. Sometimes we need the full Var node (like in the LHS of an assignment) and sometimes we need only the Name part (like in a binary operation). A helper function (get_name_node) was added to help with maintaining this distinction.
- Fixed the pytest script for the Arithmetic operations to properly use the correct node type when checking the operands of Binary/Unary operations (Uses Name instead of Var now).
- Changed the operation types from the standard types (+,-, etc...) to using the agreed-upon AST operations (ast.Add, ast.Sub, etc...). The Arithmetic operation test script has been changed to reflect this accordingly.
- Added missing source reference information for all currently existing CAST generation.
- Removed some unnecessary control characters from the control characters table in node_helper.py

Resolves #500


skema/program_analysis/CAST/python/ts2cast.py
```diff
@@ -80,6 +90,8 @@ def run(self, root) -> List[Module]:
         # Which can then contain multiple things (Handled at module visitor)
         return self.visit(root)
 
+    # TODO: node helper for ignoring comments
```
vincentraymond-ua, line 93: If you haven’t already seen it, earlier this week I merged an update to the Fortran node_helper script that adds support for removing comments

skema/program_analysis/CAST/python/ts2cast.py
```diff
@@ -147,14 +155,92 @@ def visit_expression(self, node: Node):
...
+
+        parameters = get_children_by_types(node, "parameters")[0]
+        parameters = get_non_control_children(parameters)
+        
+        # The body of the function is stored in a 'block' type node
+        body = get_children_by_types(node, "block")[0].children
+        
+
+        func_params = []
+        for node in parameters:
```
vincentraymond-ua, line 175: Be careful about overriding the symbol `node` in this manner since it is also the name of a function argument. 

This will change the value of `node` further down in the code if you try to access it again.

However, you can still use it in a list comprehension without changing the value of the top level symbol: 
```Pyhon
func_params = [node for node in parameters]  
```

vincentraymond-ua (approved):
@titomeister - This looks good to me! I've left a couple of notes, but nothing that requires immediate attention

jastier (approved):
@titomeister This looks good to me.   Approved.

skema/program_analysis/CAST/python/ts2cast.py
titomeister: Oh, that's helpful! I'll look into that. Thanks!

skema/program_analysis/CAST/python/ts2cast.py
titomeister: Oh yeah, I'll keep that in mind as that will most definitely be relevant later on. Coming up with variable names is hard..

PR #597, opened 2023-10-26
new ISA feature: integrate the variable definitions in the seed selection

ualiangzhang:
- Added a new feature to extract variable definitions from the TR mentions file to assist in seed selection.
- Fixed a bug in the equation graph structure.


Free-Quarks (approved):
LGTM!

PR #600, opened 2023-10-27
Incorporate decapodes ACSet serialization

deepsana:
- First pass on extracting and serializing the Decapodes ACSet representation.
   - Includes working example of Halfar dome PDE.
- Math Expression handles Gradient and Absolute value explicitly.
- Math Expression explicitly incorporates multiplication operations for consecutive Mrows, and handles equals operators.
- Minor bug fixes and insertion of asserts.


skema/skema-rs/mathml/src/parsers/decapodes_serialization.rs
```diff
@@ -0,0 +1,441 @@
...
+                table_counts.operation_count += 1;
+                let temp_str = format!("•{}", table_counts.operation_count);
+                let temp_variable = Variable {
+                    r#type: Type::infer,
+                    name: temp_str,
+                };
+                tables.variables.push(temp_variable.clone());
+                table_counts.variable_count += 1;
+                let tgt_idx = table_counts.variable_count;
+                let projection = ProjectionOperator {
```
Free-Quarks, line 221: You might want to add a check here on if the subtraction id unary or not and have two different flows depending. If it is a urnary subtraction the `rest[1]` call will be out of bounds and make it crash, I suspect.

Free-Quarks (approved):
LGTM, but maybe wait for Adarsh's review too, since I struggle with the low level parsing logic.

adarshp (approved):

PR #601, opened 2023-10-27
Model linespan endpoint

Free-Quarks:
This is for an added endpoint that takes in a zip file containing a python script and returns the line span the dynamics should be under. 

It currently uses an Openai LLM for this, but we will replace that with a custom local model once it's ready.

### DIFF
--- a/skema/rest/tests/test_llms.py
+++ b/skema/rest/tests/test_llms.py
@@ -40,5 +40,12 @@ def test_prompt_construction():
     # formatting the prompt with input variables
     formatted_prompt = chat_prompt.format_prompt(code=code, format_instructions = format_instructions).to_messages()
 
+    # mockes the output from the model
+    output_mock = langchain.schema.messages.AIMessage(content='```json\n{\n\t"model_function": "sir"\n}\n```',additional_kwargs={}, example=False )
+
+    parsed_output = output_parser.parse(output_mock.content)
+
+    assert isinstance(parsed_output['model_fuction'], str)
     assert isinstance(formatted_prompt[0], langchain.schema.messages.SystemMessage)
     assert isinstance(formatted_prompt[1], langchain.schema.messages.HumanMessage)
+
--- a/skema/rest/tests/test_llms.py
+++ b/skema/rest/tests/test_llms.py
@@ -1,0 +1,57 @@
+from langchain.chat_models import ChatOpenAI
+from langchain.prompts import (
+    ChatPromptTemplate,
+    SystemMessagePromptTemplate,
+    HumanMessagePromptTemplate,
+)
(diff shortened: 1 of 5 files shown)


myedibleenso (changes requested):
@Free-Quarks , have you tested this locally with the other project deps?  We're using pydantic v2 in this project.  I've found that to be incompatible with langchain's prompt templates.  To ensure skema isn't going to hit the same issue, can you please add a test that instantiates a prompt template?

myedibleenso:
Downgrading pydantic to a different major version might cause unexpected behavior.  If you run into this issue in the test, maybe you can get around using the langchain prompt template by creating your own prompt string.

Free-Quarks:
> @Free-Quarks , have you tested this locally with the other project deps? We're using pydantic v2 in this project. I've found that to be incompatible with langchain's prompt templates. To ensure skema isn't going to hit the same issue, can you please add a test that instantiates a prompt template?

Thanks for pointing this out! I am using the generic output parser from langchain, not their pydantic one, so hopefully it won't be an issue. I'm using pydantic just to simplify the final serialized json for FastAPI. But I will test it today.

Free-Quarks:
@myedibleenso So I was already running on pydantic 2.3, but I also updated to 2.4.1 and tested locally and it still works on that as well. So there shouldn't be an issue with pydantic v2!

skema/rest/tests/test_llms.py
```diff
@@ -0,0 +1,57 @@
...
+    human_message_prompt = HumanMessagePromptTemplate.from_template(human_template)
+
+    # combining the templates for a chat template
+    chat_prompt = ChatPromptTemplate.from_messages([system_message_prompt, human_message_prompt])
+
+    # formatting the prompt with input variables
+    formatted_prompt = chat_prompt.format_prompt(code=code, format_instructions = format_instructions).to_messages()
+
+    # running the model
+    output = openai(formatted_prompt)
```
myedibleenso, line 53: I don't think we'll want this API key to be available during CI (lots of calls, risk of leakage, etc.).  Can we either a) mock this part or b) test only the prompt's format/structure?

myedibleenso (changes requested):

skema/rest/tests/test_llms.py
Free-Quarks: Understandable, I'll look into that!

Free-Quarks: Okay! I am testing the prompts and mocked up the LLM response to test the parser!

myedibleenso (approved):
LGTM!  Thanks, @Free-Quarks , for adding those unit tests.

PR #602, opened 2023-10-30
Fn-Preprocessing

Free-Quarks:
This adds a util function to the REST API that takes in a function network and checks for all known bugs that can cause a deserialization error downstream. It then addresses the bugs in a generic fashion to allow serialization (although the FN is not correct, just digestible). It also outputs a log which contains every bug it has found and recording the type of bug and where the bug is inside the FN. 

I call this FN-Preprocessing on the FN's generated in the larger workflows as well. 

I also setup some error handling incase the dynamics could not be parsed in code2amr. Returning a blank/default AMR instead of a crash.


vincentraymond-ua (approved):
@Free-Quarks - Thanks for making this, looks good to me!

PR #603, opened 2023-10-30
Incorporated Decapodes Ascet serialization of Halfar dome equation 8.2 from CISM documentation

deepsana:
Incorporated Superscript of right hand side parenthesis (e.g.`)^{n+1}`) and test case for decapodes ASCet serialization of Halfar dome equation 8.2 from CISM documentation


Free-Quarks (approved):
LGTM!

PR #608, opened 2023-10-31
[execution_engine] Endpoint for parameter extraction and amr enrichment

vincentraymond-ua:
## AMR enrichment endpoint
Creates an endpoint `amr-enrich` which accepts an AMR Dict, a source file string, and a filename. The endpoint uses the execution engine to extract values for the parameters listed in the AMR and returns an enriched version with the `value` field filled in.

### Testing
Creates test_execution_server.py for unit tests relating to execution_engine endpoints

### Other changes
- Improved error handling for execution engine allows
- Fixes bug where path to Gromet not being properly resolved
- Adds additional Gromet preprocessing step before Memgraph ingestion
- Fixes floating point precision bug
-  Adds :Opi as possible node type for assignment right hand side

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,8 +84,11 @@ core = ["skema[img2mml]", "skema[isa]", "skema[tr]", "skema[metal]"]
 # see skema/img2mml/render_mml/mathpix_annotator
 annotations = ["matplotlib", "notebook"]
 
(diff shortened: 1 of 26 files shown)


Free-Quarks (approved):
LGTM! Thanks!

PR #610, opened 2023-10-31
Ensure `skema[llms]` included in docker image build

myedibleenso:
## Summary of Changes

https://github.com/ml4ai/skema/pull/601 introduced LLMs using `langchain` and the Open AI API.  The `llms` extra was included in the extra `all`, but our docker image only installs `core` to avoid bloat:  https://github.com/ml4ai/skema/blob/3af73f74ec127e0fb4c5abf4d3a8f9bd0eea3e71/Dockerfile.skema-py#L53

This caused an import error that triggered a deployment failure loop.


Free-Quarks (approved):
LGTM! Thanks Gus!

Issue #593, opened 2023-10-24
 Support Tree-sitter parsing of MATLAB conditional logic


jastier:
## Support Tree-sitter parsing of MATLAB conditional logic
Support CAST generation and validation testing of this code snippet: 
```matlab 
if x > 5
    y = 6
elseif x > 0
    y = x
else
    y = 0 
end
```

This will be used to represent MATLAB switch statement logic not covered by CAST
Related to issue #561

Issue #561, opened 2023-10-10
Support Tree-sitter MATLAB switch statement


jastier:
## Support Tree-sitter parsing and unit testing of the MATLAB switch statement
Support CAST generation and validation testing of this code snippet:

``` matlab
switch x
    case 1
        n = 1;
        j = 1;
    case {3, three, 'three'}
        n = 3;
    case {{1, 2, 3,}, {7, 8, 9}}
        n = 6;
    otherwise
        n = 0;
end
```

jastier:
PR #634 resolves this issue and is up for review.

jastier:
PR #634 has been merged into main, issue resolved and closed.

PR #613, opened 2023-11-01
MATLAB conditional logic 

jastier:
## MATLAB conditional logic added to CAST parser

This PR fixes the if statement handler in `matlab_to_cast.py`.  We now fully support `if elseif else` conditional logic of arbitrary depth in MATLAB.

- One `if` statement is always included
- Zero or more `elseif` clauses can be included
- Zero or one `else` clause can be included

## Relevant issues
Completes issue #593
Partially completes  issue #561


vincentraymond-ua (approved):
@jastier - Looks good!

PR #615, opened 2023-11-01
Pin node minor version to avoid occasional CI issue

myedibleenso:
## Summary of Changes

[Nodesource changed their Node JS binary installation process back in August](https://github.com/nodesource/distributions/pull/1626).  This seems to have finally caught up with us.

[This PR shifts to the new installation method](https://github.com/nodesource/distributions#nodejs).


myedibleenso:
FYI @Free-Quarks @vincentraymond-ua

PR #618, opened 2023-11-01
[CI] PR template and new auto-labels for PRs

myedibleenso:
## Summary of Changes

Adds a PR template for greater consistency, as well as new auto-label triggers for PRs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/summary_of_changes.md
+++ b/.github/PULL_REQUEST_TEMPLATE/summary_of_changes.md
@@ -1,0 +1,3 @@
+## Summary of Changes
+
+
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,4 @@
+# see https://github.com/actions/labeler
 documentation:
 - any: ["docs/**"]
 #  all: ["!docs/changes.md"]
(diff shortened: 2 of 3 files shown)


myedibleenso:
@vincentraymond-ua , can you help me debug this?  I'm not sure why the tests are suddenly failing after reordering a few lines in the `pyproject.toml` ...

vincentraymond-ua:
@myedibleenso - Yes, I can take a look into this too and see if I can figure anything out

vincentraymond-ua:
I need to do some more testing to be sure, but it looks like this could be a test ordering issue. It seems like `skema/img2mml/tests/**` may need to be run first.

myedibleenso:
Thanks, @vincentraymond-ua , for debugging that.  Any sense why those tests need to run first?  I'm guessing it has to do with retrieving the model:

- https://github.com/ml4ai/skema/blob/badc8a358765b41ef1d09e3097b5b8275e93bc53/skema/img2mml/tests/test_eqn2mml.py#L14

[There is probably a way we can configure pytest to run something prior to executing tests](https://docs.pytest.org/en/6.2.x/writing_plugins.html#localplugin).  For the time being, though, we should probably add a note in the `pyproject.toml`.

vincentraymond-ua:
@myedibleenso - I'm not super familiar with the img2mml code, but I believe you are correct. When not run first there is an issue locating this resource: `arxiv_im2mml_with_fonts_with_boldface_vocab.txt`

The other restriction I found was with the `test_primitive_map` unit test in `skema/program_analysis/tests`. The `build_map` assert will return False if the primitive map has already been built, which can also occur in the execution engine.

vincentraymond-ua (approved):
Looks good! Once the CI passes, this should be ready to merge.

myedibleenso:
Thanks, @vincentraymond-ua !

PR #619, opened 2023-11-01
CAST test upgrades

jastier:
## CAST test upgrades
This PR includes some test utilities for use when testing CAST output, so common tests such as an assignment or an expression can be done as single function calls.  Some debug scripts are now executables.   Minor edits and code cleanup

## What's New
- Added `skema/program_analysis/CAST/matlab/tests/utils.py` with functions for common testing tasks
- Added 3 conditional logic tests:
          `if`
          `if else`
          `if elseif else`

## Relevant issues
Related to PR #613 and #583

### DIFF
--- a/skema/program_analysis/CAST/matlab/cast_out
+++ b/skema/program_analysis/CAST/matlab/cast_out
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import json
 import sys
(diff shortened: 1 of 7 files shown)


titomeister (approved):
Looks good to me! I like the use of the helpers that you wrote to check different CAST nodes. I might have to do something like that on my end.

PR #620, opened 2023-11-06
Implemented trimming of post-processed commas and periods in MathML

deepsana:
Implemented trimming of post-processed commas and periods in MathML.


Free-Quarks (approved):
LGTM!

Issue #621, opened 2023-11-07
[MORAE] Make new codebase2amr endpoint that uses the Linespan endpoint


Free-Quarks:
This is an additional endpoint, perhaps called llm-assisted-code2amr, which will use the Linespan endpoint we currently have to sub-select the code to only the relevant part for an AMR extraction and send that through the code-snippets-2AMR pipeline. 

Some notes:
- We want this endpoint and the Linespan both available in the unified service. TA-4 is interested in using the linespan endpoint alone as well for sending things to our snippets endpoint with HMI. This is also why the output of the linespan endpoint is what it is, since they already had support for that data structure. 
- The linespan endpoint currently uses GPT3.5 for the extraction. This is temporary until we replace it with our own model that operates on function networks. A downside of the LLM model (besides response time) is that it operates on the source code itself, so it currently only operates on one code file, despite us calling it a codebase2amr, it will only work for one file in the zip until we update the model to our own. I didn't think it was worth the effort to engineer it to handle a codebase of arbitrary size, since we will be replacing it hopefully soon, but that is an option and I thought worth noting.

PR #633, opened 2023-11-07
Draft of endpoint for llm-assisted extraction

Free-Quarks:
This is the draft for the endpoint. Unfortunately I'm leaving you all with most of the work on it. 

It needs an algorithm that takes in the starting and ending line numbers and slices the code in the zip file and sends it to the snippets endpoint via the `code2fn.System` object. 

To keep the workflows.py sleek and clean, most of that fuction should probably be moved to the utils.py as well. 

Lastly, as mentioned in #621 , we are only parsing the first .py file in the zip drive until we add our own model that operates off Function Nets. So keep in mind when slicing. 

I can also finish this when I get back if needed.


vincentraymond-ua:
Closing this stub PR, will be merged in #637

PR #634, opened 2023-11-08
MATLAB switch statement support

jastier:
## MATLAB switch statement support
This PR adds Tree-sitter coverage of  MATLAB switch statements.  CI tests are also included.

``` matlab
switch x
    case 1
        n = 1;
        j = 1;
    case {3, three, 'three'}
        n = 3;
    case {{1, 2, 3,}, {7, 8, 9}}
        n = 6;
    otherwise
        n = 0;
end
```

## Relevant features
- Switch statements are translated into CAST conditional logic
- Multiple argument cases are reduced to single list inclusion test
- Case arguments may be of differing datatypes  (string, number, identifier)

## Related issues

- Resolves issue #561


skema/program_analysis/CAST/matlab/matlab_to_cast.py
```diff
@@ -578,73 +569,119 @@ def visit_do_loop_statement(self, node) -> Loop:
...
+        def get_case_body(case_node):
+            """ return the instruction block for the case """
+            block = get_first_child_by_type(case_node, "block")
+            if block:
+                return valid([self.visit(child) for child in block.children])
+            return None
+            
+        def get_model_if(case_node, identifier):
+            """ return conditional logic representing the case """
+            return ModelIf(
```
titomeister, line 623: Here, you should add the source_ref field that is missing.

skema/program_analysis/CAST/matlab/matlab_to_cast.py
```diff
@@ -578,73 +569,119 @@ def visit_do_loop_statement(self, node) -> Loop:
...
+                last.orelse = valid([self.visit(child) for child in block.children])
+
+        return model_ifs[0]
 
+    def visit_if_statement(self, node):
+        """ return a node describing if, elseif, else conditional logic"""
+
+        def conditional(conditional_node):
+            """ return a ModelIf struct for the conditional logic node. """
+            ret = ModelIf()
```
titomeister, line 652: You'll probably also need to add the source_ref that's missing here too.

skema/program_analysis/CAST/matlab/matlab_to_cast.py
```diff
@@ -78,6 +82,7 @@ def generate_cast(self) -> List[CAST]:
         """Interface for generating CAST."""
 
         # remove comments from tree before processing
+
         modules = self.run(remove_comments(self.tree.root_node))
         return [CAST([module], "matlab") for module in modules]
```
titomeister, line 87: I'm going to guess that this line would be a good place to start for resolving the extra list being wrapped around the CAST object.

vincentraymond-ua: I think this is based on the Fortran TS2CAST which can return multiple CAST modules for a single file. This is then taken into account in multi_file_ingester.py to handle creating Gromet for each module.

jastier: If possible I would like to shift this fix to the next PR

skema/program_analysis/CAST/matlab/matlab_to_cast.py
jastier: FIxed

skema/program_analysis/CAST/matlab/matlab_to_cast.py
jastier: Fixed

titomeister (approved):
Looks good!

PR #637, opened 2023-11-09
[morae] LLM assisted code->AMR endpoint + support for CHIME-SIR

### LINKED ISSUE #621: [MORAE] Make new codebase2amr endpoint that uses the Linespan endpoint

### LINKED ISSUE #628: [MORAE] Handling CHIME in code2AMR
This is an issue to track handling the CHIME model for code2AMR.

vincentraymond-ua:
## Summary of changes
Adds a new workflow endpoint to skema.rest `llm-assisted-codebase-to-pn-amr` that slices the source code based on model dynamic linespans determined by an llm. This greatly increases the accuracy of AMR generation.

Enables support for generating AMR for the CHIME-SIR model, which was previously failing with the normal `codebase-to-pn-amr` endpoint.

Adds a basic test case for testing CHIME-SIR->AMR generation.

Resolves #621
Resolves #628


myedibleenso:
@vincentraymond-ua , we're trying to avoid use of the API key in CI.  Can we mock (at least a portion of) the test involving the call to the Open AI API?

vincentraymond-ua:
@myedibleenso - Gotcha, I can come up with a workaround for the tests then that mock the output from the llm.

myedibleenso (approved):
Thanks, @vincentraymond-ua .  I think this looks like a great first version ready to be merged.  @Free-Quarks can provide some feedback when he returns.

Issue #590, opened 2023-10-20
MATLAB keywords


jastier:
## MATLAB keywords

- Create a document listing the MATLAB keywords,  Functions, and Subroutines found in the Waterloo Model.  
- Indicate which words are covered by the Tree-Sitter CAST processor.

This file is a good example of what is needed:

`skema/program_analysis/CAST/fortran/fortran_builtins.yaml `

```[tasklist]
### Tasks
```

jastier:
Resolved by PR #641 

Closed

Issue #562, opened 2023-10-10
CAST translation of MATLAB functions


jastier:
## CAST translation of MATLAB functions

Implement CAST for the following:

```matlab
function addition_demo = both(x, y)
    both = x + y;
end

y = addition_demo(1, 3);
```

### Matlab Tree-sitter syntax nodes to support:
- function_definition
- function_output
- function_arguments
- function_call  (done)

Issue #636, opened 2023-11-08
CAST handling of MATLAB matrix literal


jastier:
Add CAST coverage for the MATLAB matrix literals:

``` matlab
x = [1  2  3]
```

Note that MATLAB does not place commas between matrix values.

jastier:
Resolved by PR #[641](https://github.com/ml4ai/skema/pull/641)

Closed

Issue #635, opened 2023-11-08
Add Tree-sitter parsing support for MATLAB unary operator


jastier:
Add CAST coverage for the unary negation operator:

``` matlab
x = -6
```

Currently the negation sign is dropped when these values are read

jastier:
Resolved by PR #641 

Closed

PR #640, opened 2023-11-14
MATALAB function call and matrix integration

jastier:
## MATLAB function call, operator and matrix support, code cleanup
This PR adds support for Operators, matrix literals, and function calls.  I also got rid of some old Fortran routines that didn't work with MATLAB.  Added support for some Operator types, we now support four of them and have three remaining.  Those are going to have to be resolved by translating them to CAST loop constructs.  

## Relevant features:
- Added Operator support (Boolean, comparison, unary)
- Added function call node support
- Added n-dimensional matrix support for all primitive types
- Added new CI test cases for all of the above
- Added a file with Waterloo model Tree-sitter syntax tokens

## Bugfixes
- Fixed a bug where CAST output was getting wrapped in an extra list.
- Fixed a bug were the source_refs node wasn't getting included on certain structures.  Our CI now checks them all.

## Related issues
Resolves issues #590, #635, #636
Partially resolves issues #562


jastier:
Too many merge conflicts, I'm dropping this PR and creating a new one.

PR #641, opened 2023-11-14
MATLAB function call, operator and matrix support, code cleanup

### LINKED ISSUE #590: MATLAB keywords

jastier:
## MATLAB function call, operator and matrix support, code cleanup
This PR adds support for Operators, matrix literals, and function calls.  I also got rid of some old Fortran routines that didn't work with MATLAB.  Added support for some Operator types, we now support four of them and have three remaining.  Those are going to have to be resolved by translating them to CAST loop constructs.  

## Relevant features:
- Added Operator support (Boolean, comparison, unary)
- Added function call node support
- Added n-dimensional matrix support for all primitive types
- Added new CI test cases for all of the above
- Added a file with Waterloo model Tree-sitter syntax tokens

## Bugfixes
- Fixed a bug where CAST output was getting wrapped in an extra list.
- Fixed a bug were the source_refs node wasn't getting included on certain structures.  Our CI now checks them all.

## Related issues
Resolves #590 #635 #636
Work towards #562


skema/program_analysis/CAST/matlab/inspect
myedibleenso, line 1: Let's add developer documentation for these runnables you're adding, @jastier .  You can add this to the website (the relevant source is under `docs/dev`).

skema/program_analysis/CAST/matlab/matlab_to_cast.py
```diff
@@ -796,380 +529,206 @@ def visit_math_expression(self, node):
...
+        """ Translate MATLAB while_loop syntax node into CAST Loop node """
         """
-        Node structure
-        Do while
-        (do_loop_statement)
-            (while_statement)
-                (parenthesized_expression)
-                    (...) ...
-            (body) ...
+        SOURCE:
```
titomeister, line 640: I think at some point in the future it makes sense to remove these comments that have the code and syntax tree, or at least move them somewhere out of the way. This way the code is easier to read through.
However I think for now it's fine since I imagine it helps out with development.

jastier: Those were Tree-sitter output that I use for designing CAST structures for unimplemented syntax keywords.   I have removed them, they are easy to regenerate as I need them.

titomeister (approved):
Looks good to me!

skema/program_analysis/CAST/matlab/inspect
jastier: That is a good idea, however in this case I am no longer including the output from that script in the source code, so there is no further need for the script.   It has been removed from the repo.

PR #644, opened 2023-11-14
[Python] Function Keyword Argument and Function Call through Parameter PR

### LINKED ISSUE #592: [code2fn] SIDARTHE Keyword Argument Bug
Justin pointed another issue in the SIDARTHE model. 
This time, the problem lies when function names are used as keyword arguments, they aren't properly being passed in and cause negative wiring issues. 
This is due to some missing code to handle this in the GroMEt generation. The major pieces are in place, they just aren't being used appropriately.

titomeister:
### GroMEt generation
- Added support for handling the GroMEt primitive "_call" 
- Fixed an issue with the unpacking of a tuple from an attribute item
- Fixed an issue with packing of a tuple in function calls.
- Added some additional code to make the handling of Attributes more robust
- Fixed an issue with functions being used as keyword arguments
- Fixed an issue with port indices being too large in some wires.

### Python to CAST
- Added a table to maintain a function's arguments. This is used to determine if an argument is being called as a function. 
- Added support for generating a primitive "_call" Function Call node that's used whenever a function that's passed as a function parameter is being called.

### Other
- Added "_call" primitive to the execution engine's set of primitives in order to properly support GroMEt generation for it. 

### Testing
- Added a test file that checks that the generation and use of the "_call" primitive is correct and consistent.  

Resolves #592


skema/gromet/execution_engine/types/other.py
```diff
@@ -129,3 +129,16 @@ class Range:
 
     def exec(input: int) -> range:
         return range(input)
+
+
+class Call:
```
vincentraymond-ua, line 134: We eventually will want to stop using the primitives in `skema/gromet/execution_engine/types/` since they are artifacts left over from an outdated version of the execution engine. We have a YAML file which I believe should compile the same information in a more approachable way at `skema/gromet/primitive_map.yaml`

How large of a change would this require for the Gromet generation?

This is also something we may want to combine with python_builtins.yaml or create a home directory for these type of files.

skema/program_analysis/tests/test_fun_arg_fun_call.py
```diff
@@ -0,0 +1,323 @@
...
+    a = x(z) + y(z)
+    return a
+
+def a(f): return f + 1
+def b(f): return f + 2
+
+foo(x=a,y=b,z=1)
+    """
+
+def generate_gromet(test_file_string):
```
vincentraymond-ua, line 28: This is a minor thing, but since this function is defined in multiple test files, it may be useful to add it to `skema/utils`

skema/gromet/execution_engine/types/other.py
titomeister: Oh okay. I had forgotten about that. I think as long as we can use the same interface all that would be needed would be to update what's going on behind the scenes. That is, I just need to change the interface to pull from the correct primitive map file. I don't expect there to be any reason to explicitly change anything in the Gromet generation code itself.

skema/program_analysis/tests/test_fun_arg_fun_call.py
titomeister: Right, there's a lot of repetitive things in the test cases that I would eventually like to add there. This is another good thing to add though.

myedibleenso:
@titomeister , can you please run `python skema/program_analysis/model_coverage_report/model_coverage_report.py` locally on this branch and report overall coverage stats for our models? I just want to know if the import issue is going to reduce our coverage at all.

titomeister:
> @titomeister , can you please run `python skema/program_analysis/model_coverage_report/model_coverage_report.py` locally on this branch and report overall coverage stats for our models? I just want to know if the import issue is going to reduce our coverage at all.

I run into this error when I run that script. I'm not sure what the cause is at this time.

`/Users/ferra/.venvs/skema/lib/python3.8/site-packages/pydantic/_internal/_config.py:261: UserWarning: Valid config keys have changed in V2:
* 'schema_extra' has been renamed to 'json_schema_extra'
  warnings.warn(message, UserWarning)
Traceback (most recent call last):
  File "skema/program_analysis/model_coverage_report/model_coverage_report.py", line 31, in <module>
    from skema.rest.utils import fn_preprocessor
  File "/Users/ferra/Desktop/Work_Repos/skema/skema/rest/utils.py", line 4, in <module>
    from askem_extractions.data_model import AttributeCollection, AttributeType, Mention
ImportError: cannot import name 'Mention' from 'askem_extractions.data_model' (/Users/ferra/.venvs/skema/lib/python3.8/site-packages/askem_extractions/data_model/__init__.py)`

vincentraymond-ua:
@titomeister - Could you reinstall Skema and see if that error goes away? I have a feeling that this is something that may have been added since you last installed the project.

jastier:
This is a bug in my code I believe.   I am now putting up a PR that should fix it.

titomeister:
> @titomeister - Could you reinstall Skema and see if that error goes away? I have a feeling that this is something that may have been added since you last installed the project.

Thank you @vincentraymond-ua, that resolved that problem! Working on some stuff with the report it generated now.

vincentraymond-ua (approved):
@titomeister - Looks good! Sorry for taking so long to re-review it.

titomeister:
> @titomeister - Looks good! Sorry for taking so long to re-review it.

No worries! Thank you for looking at it.

PR #650, opened 2023-11-15
[rest] Adding SKEMA_GRAPH_DB_HOST and SKEMA_GRAPH_DB_PORT environment variables

vincentraymond-ua:
## Adding environmental variables
Adds the `SKEMA_GRAPH_DB_HOST` and `SKEMA_GRAPH_DB_PORT` environmental variables to skema/rest. 

Updates the execution engine to use these values to connect to Memgraph.

See the corresponding PR to set these values in the ASKEM-TA1-DockerVM repo:
https://github.com/ml4ai/ASKEM-TA1-DockerVM/pull/44


Free-Quarks (approved):
LGTM!

myedibleenso (changes requested):
@vincentraymond-ua , this is a great start, but we will also need to modify how the `skema-rs` REST API receives this information.

Free-Quarks:
The information is passed as a command line argument on the docker side entrypoint for RUST. So this should still work for both services, since the environment variable is also set on the docker side.

myedibleenso:
>The information is passed as a command line argument on the docker side entrypoint for RUST service. So this should still work for both services, since the environment variable is also set on the docker side.

Yeah, I guess so.

Free-Quarks:
I agree it should be an environment variable and is a left over way of doing things before dockerized everything

myedibleenso:
Actually, I take that back.  I think a port is assumed: https://docs.rs/rsmgclient/latest/rsmgclient/struct.ConnectParams.html

Could we at least optionally specify the db port in the skema-rs runnable with the standard bolt-like default?

myedibleenso:
```rust
let connect_params = ConnectParams { 
         host: Some(host.to_string()), 
         port: Some(port.to_int()),
         ..Default::default() 
     }; 
```

... and:

```rust

#[derive(Debug)]
   pub struct Config {
      pub db_host: String,
      // FIXME: how to set a default (ex. 7687) here?
      pub db_port: int
   }


```

myedibleenso:
^ idk if my Rust is correct (probably not), but hopefully it at least conveys the idea.

Free-Quarks:
Yeah I think I get what you are saying. (looping in @adarshp to double check as he wrote the code) We have it setup to get the port from the command line argument, but when we hit Memgraph we don't specify it each time and assume it is static, but we could specify it in the way you are showing. I could add that to this PR.

myedibleenso:
I see a lot of use of `host` in a bunch of places, so probably doing this would involving touching many lines of code...

myedibleenso:
I guess I can try to deal with it on the deployment side.  I'll drop it for now...

myedibleenso (commented):
Just noting that a DB port cannot currently be passed on the Rust side.

myedibleenso (approved):

myedibleenso:
@vincentraymond-ua , do you these env variables need to be noted in our documentation, or is it sufficient to update the dockerVM repo?

vincentraymond-ua:
@myedibleenso - Thats a good question, it may be valuable to have these described in the documentation, perhaps in the dev environment page: https://ml4ai.github.io/skema/dev/env/

PR #652, opened 2023-11-15
[MORAE] Multi-file support for linespan and code2amr slicing

Free-Quarks:
This PR fixed a small bug about making sure to only slice files that are in .py extensions.
It also added support for multi-file systems in both the linespan endpoint and the llm-assisted-codebase-to-pn-amr endpoint. 

Some notes about the implementation, there is an assumption that zip files are walked over the same way each time and we are still only returning one AMR, to be consistent with the other code2amr endpoint. In the case we get multiple AMR's, (unpredictable LLM stuff) we will only return the AMR which has the most "states", which we use as a proxy for completeness. Default is the first amr, if say all amr's are empty.


myedibleenso:
I think we can close this in favor of #658 .

PR #656, opened 2023-11-17
Matrix and list handling bugfixes, improved testing

jastier:
## Matrix and list handling bugfixes, improved testing
This PR fixes an issue with matrix handling where nested empty matrices would be represented as a single empty matrix.  We now show the empty matrices in their correct nested positions.  Also some better CI testing.  Instead of picking off individual fields to test, the only input now is another instance of the same class the way it should look, and the code will check it in one go.  Added function call tests with operators and other functions as arguments.

## Relevant features
- Correct representation of empty matrices.
- CI testing with class instances.  More test cases.

 ## Related issues
Partially addresses #562


titomeister (approved):
Looks good to me!

PR #658, opened 2023-11-17
[memgraph/morae]  env-based memgraph connection changes and multi-file system support for linespans

myedibleenso:
## Summary of Changes

- (@Free-Quarks ) Only slice files that are in .py extensions
- (@Free-Quarks ) Adds support for multi-file systems in both the linespan endpoint and the llm-assisted-codebase-to-pn-amr endpoint
- Uses `SKEMA_GRAPH_DB_HOST` and `SKEMA_GRAPH_DB_PORT` environment variables in Rust
- Uses `SKEMA_GRAPH_DB_HOST` and `SKEMA_GRAPH_DB_PORT` environment variables in CI test to connect to Memgraph service
- Renames PR template so that it is picked up when opening pull requests.

### Notes 

`SKEMA_GRAPH_DB_HOST` must cont contain the protocol (ex. `bolt://`) when used with rsmgclient (i.e., `my-url.com` NOT `bolt://my-url.com`).

Regarding implementation decisions for supporting multi-file systems (from @Free-Quarks ), ...
>there is an assumption that zip files are walked over the same way each time and we are still only returning one AMR, to be consistent with the other code2amr endpoint. In the case we get multiple AMR graphs (ex. nondeterministic LLM behavior), we will only return the AMR which has the most "states", which we use as a proxy for completeness. If all AMR graphs are empty, we return the first AMR graph.


Free-Quarks (approved):
LGTM! I also tested it locally with default values for the environment variables and it worked perfectly for me on the Rust side!

PR #672, opened 2023-11-20
[execution_engine] Switching to neo4j driver 

vincentraymond-ua:
## Summary of Changes

- Switches Execution Engine to neo4j driver from gqlalchemy Memgraph driver. 
- Now uploads Gromet to Memgrpah database using skema-rs service call instead of using a local subprocess. 
- Execution Engine no longer clears Memgraph state before execution. It is now capable of selectively operating on a single model and will delete the model using a skema-rs call after execution finishes.
- Adds `SKEMA_GRAPH_DB_PROTO` environment variable


Free-Quarks (approved):
LGTM! 

So the protocol defaults to bolt? That's cool we are running on bolt now for memgraph

myedibleenso (approved):
Nice work, @vincentraymond-ua !  We're free of that subprocess call.  I noted the addition of the `SKEMA_GRAPH_DB_PROTO` env variable.  It may take a few tweaks on the deployment side to get this working with our setup after this is merged.

skema/gromet/execution_engine/query_runner.py
```diff
@@ -1,19 +1,23 @@
 import yaml
 import traceback
 from pathlib import Path
-from gqlalchemy import Memgraph
+
+from neo4j import GraphDatabase
```
myedibleenso, line 5: Ack!  I wasn't thorough enough in my review.  It looks like we haven't added this to `pyproject.toml`.  We may also want to remove the old dep we're replacing here.

vincentraymond-ua: Ah you are right, let me fix that.

PR #692, opened 2023-11-27
[rest] Fixing typo in environment-variables endpoint

vincentraymond-ua:
## Summary of Changes
Fixes typo in environment-variables endpoint that was causing issues with deployment


myedibleenso:
This makes me think that we should enforce a policy where each new endpoint must have at least one unit test.

myedibleenso (approved):
We should add a unit test for this endpiont.

PR #693, opened 2023-11-27
Changes to docker build settings

myedibleenso:
## Summary of Changes

- Always attempt to pull when building a docker image (to avoid using an outdated base image)
- Disable cache during image builds (to help surface potential problems)
- Bump versions of docker actions


vincentraymond-ua (approved):
@myedibleenso - Looks good to me!

PR #694, opened 2023-11-28
Model Coverage Regression Testing PR

### LINKED ISSUE #677: [code2fn/testing] Model Coverage Regression Tests
Using the code2fn coverage report generator, create a test (or tests?) that use the results of the report generator to determine if we have a regression in line number coverage. That is, we use the line number coverage information to create a test that helps us determine if we lost line coverage or not.

titomeister:
This PR adds some modifications and a test case to help us determine when we've had a regression in model coverage. This is part of the effort to develop more tests to maintain stability within the Code2FN pipeline.

## Summary of Changes
- Adds a test script (test_model_coverage.py) that checks the coverage against all the current models we support, it also checks to make sure that whenever we run the test pipeline that we generate reports for all the models that we support.
- Modified model_coverage_report.py to output an additional JSON file that contains the 'raw' model coverage information (i.e. the supported/total lines) that is used by the test script to check for regressions in model coverage.

Resolves #677

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,11 +41,11 @@ jobs:
     steps:
     # Setup docker
     - name: Set up QEMU
-      uses: docker/setup-qemu-action@v2
+      uses: docker/setup-qemu-action@v3
     # for multi-arch builds (ex. ARM 64)
     - name: Set up Docker Buildx
       id: buildx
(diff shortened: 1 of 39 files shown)


vincentraymond-ua (approved):
@titomeister - Looks great, thank you Tito!

PR #704, opened 2023-11-29
Rust neo4rs Driver Switch

Free-Quarks:
## Summary of Changes
This should update the Rust side to using neo4rs neo4j drivers from the rsmgclient client drivers for interacting with memgraph. 

Needed a fair bit of refactoring in the model_extraction.rs file in order to accommodate this change, so I suggest most review effort should be there.


adarshp (approved):

skema/skema-rs/skema/src/config.rs
```diff
@@ -1,36 +1,57 @@
+use neo4rs::*;
 use std::env;
-use rsmgclient::{ConnectParams};
+// NOTE: this is deprecated
```
myedibleenso, line 3: I think this can be removed, @Free-Quarks .  I had such a note on the methods associated with the rsmgclient connection.  Now that those methods have been removed/the migration to neo4rs is complete, this comment is no longer relevant.

myedibleenso (commented):
Nice work, @Free-Quarks !  From why I can tell, this all looks good (tests are passing, etc.).  I suggest removing the comment from config.rs that I noted to avoid confusion.

skema/skema-rs/skema/src/config.rs
Free-Quarks: Sounds good! Done!

Thanks!

PR #708, opened 2023-11-30
Update reader to processors 9.0.0-RC2

kwalcock:
@enoriega, this shows how the main branch can be changed to use the updated processors.  It was also necessary to update pdf2txt, which also depends on processors.


enoriega:
@kwalcock thank you! Testing ...

PR #722, opened 2023-12-07
[fortran] Further bug fixes and speed improvements

### LINKED ISSUE #703: [gromet] Resolve Tuple Return issue
There is a bug in the pack_tuple_return function in to_gromet_pass.py.

The error is occurring because of am empty pof somewhere:
``` bash
last_pof = len(gromet_fn.pof)
TypeError: object of type 'NoneType' has no len()
```

### LINKED ISSUE #721: [fortran] Add support for derived type member expression calls

### LINKED ISSUE #724: [fortran] Fix strings that should be name nodes
There are a couple of instances where we are using raw strings to reference the name of CAST objects when we should be using Name ASTNodes instead. 

One example of this is the Attribute.attr field.

vincentraymond-ua:
### Fortran->CAST efficiency updates
- Using itertools.accumulate to generate overlapping sums in NodeHelper class
- Creating a single dummy source reference node instead of creating a new instance every time we come across a missing source reference.

### CAST->Gromet efficiency updates
- Updates the port_id generation logic 

### Fortran->CAST
- Adds `get_children_except_types` function to node_helper.py, which is the inverse of `get_children_by_types` function.
- Adds `generate_cast_body` function to TS2CAST which handles the pattern of visiting every node in a List. 
- Adds support for unary operators in expressions (a = +b)
- Adds basic support for print statements
- Adds support for function derived type member expressions (b = a%function())
- Refactors the visitor for the if conditional to increase accuracy and reduce code
- Refactors return logic to properly differentiate subroutine and function
### Related issues

Resolves #724  Resolves #721 Resolves #703


titomeister (approved):
Looks good to me! Sorry it took so long. 
I will probably have to borrow some logic from the variable context as it looks like it's more robust now.

PR #732, opened 2023-12-15
Gromet Wiring Detector PR

### LINKED ISSUE #697: [code2FN] Script for GroMEt Wiring Diagnosis
Create a script to programmatically detect missing wires in Gromet.
We can start with some target files
- A gromet with a wire that is intentionally modified (i.e. points to nowhere)
- A gromet with a wire that is intentionally removed (probably harder)

titomeister:
This PR adds a new script `skema/program_analysis/gromet_wire_diagnosis.py` that can be used to do some simple analysis and error detecting in the wires of GroMEt FNs. It currently checks the ports of all types of wires, and detects whether the ports are out of bounds (in either negative or positive indices) within their respective port tables. It also attempts to find the most relevant SourceCodeReference metadata that is associated with the wires and displays the line number information contained within it. 

## Summary of Changes
- Adds `skema/program_analysis/gromet_wire_diagnosis.py` script
- Modifies `skema/program_analysis/JSON2GroMEt/json2gromet.py` script so that it can ingest newer GroMEt JSON that uses the updated Gromet metadata fields.
- Fixes a small issue with the incorrect SourceCodeReference metadata type being used in `skema/program_analysis/CAST2FN/ann_cast/to_gromet_pass.py`
- Adds a test script `skema/program_analysis/tests/test_wiring_diagnosis.py` that can test the consistency of the individual wire checker utility without needing a GroMEt JSON. 

### Potential Next steps
- Need to determine more things that can be easily analyzed in a GroMEt JSON 
- Come up with a more robust way of determining what line numbers go with the wires.
- Determine what to do with the Metadata fields that don't have an "is_metadatum" field attached to them. (NOTE: A solution to this has been currently proposed.)

Resolves #697


vincentraymond-ua:
@titomeister - You mentioned Metadata types that don't have an is_metadatum field. These should be handled by these lines https://github.com/ml4ai/skema/blob/d5f4c61618e5d9d1389d1430efb3f10f6e0e25bf/skema/program_analysis/JSON2GroMEt/json2gromet.py#L52-L68
. If this no longer works though let me know and we can figure something else out.

titomeister:
@vincentraymond-ua So the issue I was running into is that the code would break on the if-else statement that is directly before this chunk of code.

https://github.com/ml4ai/skema/blob/d5f4c61618e5d9d1389d1430efb3f10f6e0e25bf/skema/program_analysis/JSON2GroMEt/json2gromet.py#L47-50

Currently there is temporary handling for the metadata types that don't have an is_metadatum field, but I know there's something better. Whether that involves just adding the is_metadatum field to those metadata types or something else.

vincentraymond-ua (approved):
@titomeister - Looks good! Eventually it would be nice to integrate with Justin's [fn_preprocessor](https://github.com/ml4ai/skema/blob/398bfdf641b6a191d3da4c8e4685a3b8787025f0/skema/rest/utils.py#L11) so that this shows up in the model coverage reports as well.

PR #739, opened 2023-12-22
[fortran] CISM coverage updates

### LINKED ISSUE #700: [gromet] Gromet only supports finding imported functions for Python
Currently the Gromet pipeline only supports checking if a function is an imported function for Python source code. If a Fortran/Matlab source does the same, it will crash during the Gromet ingestion stage.

Current implementation:
``` Python
def find_func_in_module(module_name, func_name):
    """When a module is imported using
    'from x import *', all the functions for module x
    then become bound to the namespace but don't have their attributes attached to the module x
    That is, they're called using just their name as opposed to x.func_name().
    This function attempts to find a function in such a module by importing it and then
    using 'dir' to determine if func_name is within module_name's list of functions
    """
    # TODO: Metapath finder, find_spec
    # TODO: cache the module objects to look back so we dont do multiple imports in a map
    import os
    import sys

    sys.path.append(os.getcwd())
    module_import = importlib.import_module(module_name)

    funcs = list(dir(module_import))
    return func_name in funcs
```

### LINKED ISSUE #719: [gromet] Support empty body for if condition
There are a few cases in CISM where there are legitimate if conditionals with empty bodies:

```
   if (model%options%whichdycore /= DYCORE_GLIDE) then   ! glam/glissade dycore
!!       call coordsystem_allocate(model%general%ice_grid, model%basalproc%Hwater)
!!       allocate(model%basalproc%u (ewn-1,nsn-1,model%basalproc%tnodes)); model%basalproc%u=41.0d3
!!       allocate(model%basalproc%etill (ewn-1,nsn-1,model%basalproc%tnodes));model%basalproc%etill=0.5d0
    endif
```

This case doesn't actually have an effect on the data overall data flow, but if the expression calls a side-effecting function, it could.

vincentraymond-ua:
## Tree-Sitter Parsers
- Adds a new `--ci` flag to build_parsers.py so that the .so is only copied to site packages when running on CI. When running locally this can cause missing import issues.
- Updates Github workflow to add --ci flag to build_parsers call.

## Preprocessor
- Updates arguments to GCC invocation to explicitly specify the source language. This prevents GCC from mixing up C style comments and Fortran concatenation (//) 

## TS2CAST
- Updates generate_cast_body to add a no_op to output List if it is empty.
- Updates Loop visitor to use generate_cast_body to prevent Null values from appearing in CAST.

## CAST->GROMET
- Adds error handling to find_func_in_module function to prevent it from crashing on Fortran source code.

### Related issues
Resolves #700 
Resolves #719


titomeister (approved):
This looks good to me! Sorry it took me so long to review.

PR #741, opened 2023-12-22
[REST] Refactor for improved consistency

myedibleenso:
## Summary of Changes

- Explicitly return a status code of 200 for calls to /version.
- Support head calls to /version.
- Changed ISA PUT request to a POST
- Renamed file1 and file2 to `mml1` and `mml2` (in the REST API, these will **always** be "file" contents (i.e., MML).
- Added ISA test case data to `isa/data.py` for ease of use between ISA tests and API docs.
- Refactor status codes to use constants for better consistency.
- Temporarily disable ISA router (deployment debugging)

### DIFF
--- a/skema/img2mml/eqn2mml.py
+++ b/skema/img2mml/eqn2mml.py
@@ -7,7 +7,7 @@ node data_generation/mathjax_server.js
 
 from typing import Text
 from typing_extensions import Annotated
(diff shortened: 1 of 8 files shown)


vincentraymond-ua (approved):
@myedibleenso - LGTM!

PR #745, opened 2024-01-09
Python tree-sitter to CAST porting: Loops

### LINKED ISSUE #498: Python tree sitter: Loops

- Getting end-to-end (Source to GroMEt) generation for loop structures (For and While loops)

### LINKED ISSUE #740: [code2FN] Tree-Sitter to CAST Missing Identifiers
The Python Tree-Sitter to CAST implementation is missing identifiers for certain elements.
In particular, the very first line of the source file will have its identifiers missing.

We're thinking it's a bug with the `get_identifier()` function in `node_helper.py`. Something is off about the indexing.
When we add an additional blank line at the top of the source file, then we get the identifiers back.

The simplest observable example is the code.
`x = 2`

With no additional blank lines above it.

titomeister:
This PR introduces support for generating CAST for Loops (for/while) using tree-sitter, as part of the ongoing effort to port over the Python AST to CAST generation to using tree-sitter.

### Python Tree Sitter
- Added support for generating CAST Loop nodes using Python tree sitter.
- Specifically, we added support for Python's For and While loop CAST generation. 
- Added some support for tree-sitter "patterns: "list_pattern", "tuple_pattern", "list_pattern". These are used primarily in the For loop syntax as the item we're iterating over.
- Also added support for generating Iterator function calls, as used by the Python For Loops.
- Updated the CAST to AGraph visualizer to better visualize tuples.

### Testing 

- Added some small unit tests for loops (for/while) to determine consistency. 
- Added a small unit test for detecting missing identifiers in the first line of a Python program.

### Other Fixes
- Fixes an issue with missing identifier names when they appeared on the first line of the Python program. This was done by adding some additional handling that is specific to the first line of the program. 

Resolves #498 
Resolves #740

### DIFF
--- a/skema/program_analysis/CAST/python/node_helper.py
+++ b/skema/program_analysis/CAST/python/node_helper.py
@@ -1,3 +1,4 @@
+import itertools
 from typing import List, Dict
 from skema.program_analysis.CAST2FN.model.cast import SourceRef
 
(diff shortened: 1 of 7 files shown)


vincentraymond-ua (approved):
@titomeister - Looks good!

PR #753, opened 2024-01-16
Gromet While Loop Wiring Bug PR

### LINKED ISSUE #728: [gromet] Fix while loop wiring bug
When the expression of a while loop is a function call, the pipeline fails to produce Gromet. 

```Python
def return_true():
    return True

while(return_true()):
    print("TEST")
```

titomeister:
This PR primarily fixes a crash in the Gromet generation when there's a While loop condition consists only of a function call. 
It also fixes another minor wiring bug related to LiteralValues

### Summary of Changes
- Fixed Gromet generation crash for While loop condition of a function call.
- Fixes a missing wiring bug when a function returns a plain LiteralValue.
- Adds some small unit tests to maintain consistency.

Resolves #728


vincentraymond-ua (approved):
@titomeister - Looks good!

PR #758, opened 2024-01-18
[code2fn] Additional validation for code2fn endpoints

### LINKED ISSUE #690: [MORAE] Ignore hidden files in zip directories
We should add a prefix filter, in addition to the current suffix one, so we don't try to analyze hidden files and artifacts that come from zipping files on a Mac.

vincentraymond-ua:
## Summary of Changes

- Refactors validation handler in system2gromet to support more complex cases.
- Adds check for files in hidden directories to code2fn endpoints.
- Adds check for MACOS artifacts to code2fn endpoints.
- Generates logging for new validation checks to stdout and Gromet. 
- Adds a test case for validating hidden files and MACOS artifacts.
- Fixes bug where file could be removed from system.files buts not system.blobs
### Related issues
#690 

Resolves #690

### DIFF
--- a/skema/rest/llm_proxy.py
+++ b/skema/rest/llm_proxy.py
@@ -64,6 +64,13 @@ async def get_lines_of_model(zip_file: UploadFile = File()) -> List[Dynamics]:
         for file in zip.namelist():
             file_obj = Path(file)
             if file_obj.suffix in [".py"]:
+                # Skip file if located in a hidden directory or MACOSX artifact
+                for parent in file_obj.parents:
+                    if parent.name == "_MACOSX":
+                        continue
+                    elif parent.name.startswith("."):
+                        continue
+                
                 files.append(file)
                 blobs.append(zip.open(file).read().decode("utf-8"))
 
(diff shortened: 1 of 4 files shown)


titomeister:
@vincentraymond-ua does this resolve any issues on the kanban board? I didn't see any in the PR description but I wanted to double check.

vincentraymond-ua:
Yep, just updated the description with the issue!

Free-Quarks:
Hi @vincentraymond-ua 

So I think you are write that since the code2fn in called in the other two endpoints that it will handle the hidden files without crashing. But we might be able to get some more performance on those endpoints if we filter the files at the beginning in the same was you have implemented for code2fn, especially in the LLM case since each file we ignore is one less call to the LLM. 

So namely updating the two following locations: 
https://github.com/ml4ai/skema/blob/b089c47e4db95df8e81ba88065123933f79f6a49/skema/rest/workflows.py#L270-L275

https://github.com/ml4ai/skema/blob/b089c47e4db95df8e81ba88065123933f79f6a49/skema/rest/llm_proxy.py#L63-L68

I'm hoping it wouldn't be much more work if that's okay?

titomeister (approved):
Looks good to me!

Free-Quarks:
This looks good to me too! We can merge this and I'll make the CI tests for it in a later branch / PR.

Free-Quarks (approved):

Issue #778, opened 2024-01-30
[ISA] Code and Equation Alignment


Free-Quarks:
This is to add a new workflow endpoint to align equations from papers and the corresponding equations in code.

Free-Quarks:
Completed with PR #781

PR #781, opened 2024-02-04
[ISA] A new workflow endpoint to align equations from papers and the corresponding equations in code

ualiangzhang:
## Summary of Changes
This is to add a new workflow endpoint to align equations from papers and the corresponding equations in code.

## Solve #778


Free-Quarks (approved):
LGTM!

PR #803, opened 2024-02-16
Python tree-sitter to CAST Porting: Comprehensions and Lambdas

### LINKED ISSUE #779: Python tree sitter: Comprehensions & Lambdas

This issue is to track support for Comprehensions (List, Dict) and Lambda expressions.
We group these two idioms together because the generation solution for these are all very similar.

```[tasklist]
### Tasks
- [x] CAST support for List/Dict Comprehensions
- [x] CAST support for Lambdas
- [x] CAST unit tests for List/Dict Comprehensions
- [x] CAST unit tests for Lambdas
```

titomeister:
This PR introduces support for generating CAST for list comprehensions, dictionary comprehensions, and lambda functions. This is uing tree sitter, as part of the ongoing effort to port over the Python AST to CAST generation to use tree-sitter.

## Summary of Changes
- Adds handlers to `python/ts2cast.py` to support list comprehensions, dictionary comprehensions, and lambda expressions. It adds comprehensions with multiple loops and conditionals, and basic lambda support.
- Adds additional handling for different types of conditionals that are used in loops/if statements.
- Adds handling for the Modulo operator (%).
- Adds test scripts `test_comprehension_lambda_cast.py` and `test_adv_comprehension_lambda_cast.py` with unit tests to maintain consistency.
- Fixes some inconsistency with CASTLiteralValue types being used incorrectly. 

### Related issues
- Eventually support will need to be added for lambdas that use variables outside of its own scope.

Resolves #779


myedibleenso:
@titomeister ,  is this ready for review?

vincentraymond-ua (approved):
@titomeister - Looks good!

PR #826, opened 2024-02-27
Python tree-sitter to CAST Porting: Imports

### LINKED ISSUE #804: Python tree sitter: Imports

This issue is to track support for Imports. Add handling for imports for Python tree-sitter, and also create any appropriate unit tests. 

```[tasklist]
### Tasks
- [ ] CAST Generation for Imports
- [ ] Initial Support for CAST Generation Attribute
- [ ] Unit Tests for Imports
- [ ] Unit Tests for Attribute (probably weaved with import unit tests for now)
```

titomeister:
This PR introduces support for generating CAST import statements. This is using tree sitter, as part of the ongoing effort to port over the Python AST to CAST generation to use tree-sitter.

### Summary of Changes

- Adds handlers to python/ts2cast.py to support CAST generation for import statements.
- Adds basic CAST generation for the attribute idiom. In order to support function calls from imports.
- Adds test script test_import_cast.py with unit tests to maintain consistency.
- Added 'support' for yields and assert statements. The current generation is to passthrough them, so nothing important gets generated from these.

### Related issues
- 

Resolves #804


titomeister:
Requires some tweaks to the tests to fix a goofy issue of Var vs Name in Functions. Converting back to draft.

vincentraymond-ua (approved):
@titomeister - Looks good!

PR #834, opened 2024-02-29
[fortran] Fix for function declarations in derived types

vincentraymond-ua:
## Summary of Changes
There was a bug discovered in the CTSM source PhotosynthesisMod.F90 where a function is declared in a derived type but defined in the outer module. This PR updates the variable context to support this case. 

Additionally resolves a bug where function declarations and definitions can have different case sensitivity.

### Related issues

Resolves ???


titomeister (approved):
Looks good to me! I noticed some TODOs in the comments for the changed files. I figured I would point them out in case you need to resolve these for this PR, or more likely they're for later?

vincentraymond-ua:
@titomeister - I've updated the branch to remove the unneeded todos. Some were resolved by this fix and others are just left over things from previously.

PR #838, opened 2024-03-01
[Equations] Add the equation extraction service

ualiangzhang:
## Summary of Changes
Extract equation images from a PDF and call GPT service to check if the image contains equations.


Free-Quarks (approved):
LGTM!

PR #848, opened 2024-03-11
[CI] Pinning tree-sitter to 0.20.4

vincentraymond-ua:
## Summary of Changes
On March 10th the tree-sitter Python library updated to version 0.21.0. This version breaks many of our unit tests due to changes in language initialization. 

This PR pins the version to 0.20.4, the last working version, so that the unit tests work again. 

### Related issues

Resolves ???


myedibleenso (approved):
Thanks, @vincentraymond-ua !

PR #850, opened 2024-03-12
Python tree-sitter to CAST porting: Classes

### LINKED ISSUE #805: Python tree sitter: Classes

This issue is to track support for Classes. In CAST (and Gromet) we refer to classes by the more general Record type.

```[tasklist]
### Tasks
- [ ] Tree sitter to CAST generation for Classes
- [ ] Unit Tests
```

titomeister:
This PR introduces support for generating CAST for Python Classes. In CAST the translation for these is RecordDef. This is part of the ongoing effort of porting over the Python to CAST generation to use tree-sitter instead of the Python AST. 

### Summary of Changes
- Adds support in python/ts2cast.py for generating RecordDef CAST nodes, which represent Python classes.
- Adds test script test_record_cast.py for maintaining consistency with basic examples. 
- Adds a handler of CAST Name nodes for when they're being used in Operators or returns (get_operand_node). This is to make sure the correct CAST node is being used in these at all times.
- Very minor code cleanup.

Resolves #805

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies=[
     "dill==0.3.7",
     "networkx==2.8.8",
     "PyYAML",
-    "tree-sitter",
+    "tree-sitter==0.20.4",
     "neo4j==5.14.1",
     "requests",
     "beautifulsoup4", # used to remove comments etc from pMML before sending to MORAE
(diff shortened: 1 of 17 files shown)


vincentraymond-ua:
@titomeister - Looks good for the most part, but it looks like one of the function test cases is failing.

titomeister:
@vincentraymond-ua Fixed! Can you re-review when you get some time? Thanks!

vincentraymond-ua (approved):
Looks good!

PR #879, opened 2024-04-11
[CI] Pinning tree-sitter-cpp version to fix CI errors

vincentraymond-ua:
## Summary of Changes
A recent update to the tree-sitter-cpp repository has broken compatibility with the tree-sitter version we are using.  This PR pins the tree-sitter-cpp version to the last working commit.


skema/program_analysis/tree_sitter_parsers/languages.yaml
```diff
@@ -9,6 +9,7 @@ c:
 cpp:
   tree_sitter_name: tree-sitter-cpp
   clone_url: https://github.com/tree-sitter/tree-sitter-cpp.git
+  commit_sha: e0c1678
```
myedibleenso, line 12: Thanks for narrowing this down, @vincentraymond-ua .

myedibleenso (approved):

PR #881, opened 2024-04-19
[data] Refactors skema to fetch artifacts locally instead of from artifacts.askem.lum.ai

vincentraymond-ua:
## Summary of Changes
Refactors skema codebase to fetch data artifacts locally instead of from artifacts.askem.lum.ai.

### Changes
- skema.data submodule created to store local artifacts
- model zip archives now stored locally under skema/data
- img2mml model is stored in a huggingface repository and is downloaded at runtime.
-  huggingface_hub added as a dependency
- `deploy.yml` workflow removed 
- Old `.drone.yml` CI file removed
- Updated documentation to account for these changes


myedibleenso (approved):
Thank you, @vincentraymond-ua !

PR #888, opened 2024-04-24
[ci] Quality of life updates

### LINKED ISSUE #882: Replace defaults for `SKEMA_RS_ADDESS` and `SKEMA_MATHJAX_ADDRESS`
Since the deployment was taken down, we should replace the defaults for both  `SKEMA_RS_ADDESS` and `SKEMA_MATHJAX_ADDRESS` in [`proxies.py`](https://github.com/ml4ai/skema/blob/6cda5982af6525d5e52241f9e7d8a6ee4f2dcba1/skema/rest/proxies.py)

### LINKED ISSUE #883: [CI] `test_post_eqns_to_mets_mathml_latex` should use local mathjax service
All CI tests were meant to use "local" services, but it seems we missed launching the necessary CI service for `test_post_eqns_to_mets_mathml_latex` and it had been using the deployed mathjax service.

### LINKED ISSUE #885: [ci] Quality of life updates to model coverage  report
There are a few small enhancements we can make to make reading the model coverage reports a bit easier:
1. Change color of the Can Ingest field to match if a file can be ingested. (Green for check, Red for x)
2. Change the name of the Can Ingest field to make it clear what is being ingested (Can Ingest -> Can Produce Gromet)

vincentraymond-ua:
## Summary of Changes

- Makes the `Can Produce` field in the model coverage reports clearer by specifying which target is being generated.
- Adds color to "✓" and "✗" characters to make it clearer at a glance if a particular file could be ingested
- Updated the default values of Environment Variables in skema.rest to remove dependency on Lum.AI deployment
- Adds mathjax server to ci to support tests that previously relied on lum deployment
### Related issues
Resolves #882 
Resolves #883 
Resolves #885


myedibleenso (approved):
Thanks, @vincentraymond-ua !  I made a minor change to avoid dependency drift.

